PR #139, opened 2015-01-07
Removed reference to subfolders which do not exist

tuttinator:
Hi,

I ran into the same issue as datawrapper/datawrapper#135 and found that by removing the Makefile reference to **/* for chart.base I could get `make` to build. I ran into this issue on OS X 10.10 and Ubuntu 14.04, but I'm unsure if there are other versions of make this isn't a problem for.

Hope this helps.


gka:
Merged this into develop (which goes into master soon)

PR #165, opened 2015-06-22
Fix reload function

j0k3r:
This will fix an error after subscribed:

```
Uncaught ReferenceError: reload is not defined
(anonymous function) @ initializeSignUp.js?v=1.9.5:71
```


j0k3r:
Thanks for the feedback guys, keep up the good work :trollface: :expressionless:

PR #166, opened 2015-08-03
visualization eventsystem

plusgut:
previously it crashed when evt.data was an object, now it is typesecure


gka:
makes sense. merging. thanks!

plusgut:
you're welcome, thanks for merging

PR #167, opened 2015-08-20
Convert _ to - when building culture URL

wulff:
When building URLs to the `globalize,culture.LOCALE.js` files we need to convert "_" to "-" to avoid a 404 error (similar to the changes in 4a0c3ef and 1ace1cb).


gka:
Indeed. Good catch.

PR #168, opened 2015-08-25
Avoid infinite loop when loading globalize.culture.*.js

wulff:
The fix in my previous pull request (#167) was too specific and resulted in an infinite loop when trying to load the `globalise.culture.*.js` files:

screenshot 2015-08-21 at 14 49 10

This patch tries to remedy the situation.


gka:
yes, makes sense.

PR #169, opened 2015-08-25
Removed unnecessary substr() call

wulff:
It doesn't look like this `substr()` call is necessary.


gka:
this is a very sensitive piece of code, not sure why we need to touch it at all.

gka:
ah, yes. guess it makes sense to remove the substr..

PR #171, opened 2015-08-26
Added Danish localization

wulff:
The first full version of the Danish localization.


mirkolorenz:
Thanks for the localization - did you check Crowdin, where the bulk of the language translations for Datawrapper lives? Currently the translation status for Danish there is on 8%. This is by far the simplest way to add the translation to Danish plus keep it updated over time.

wulff:
I didn't check Crowdin, to be honest…

I'll add the localized strings over there instead :-)

davidkokkelink:
@wulff @mirkolorenz Not necessarry! We don't use CrowdIn anymore in the upcoming release, so this works very well for us ! Thank you! :)

PR #196, opened 2016-05-09
Fix nonexistent links in invitation email

Kirkman:
The invitation hooks had `invite_link` instead of `invitation_link` which was preventing the invitation emails from containing a clickable URL.


gka:
thanks, and sorry for the delay! merging...

PR #216, opened 2017-01-11
Sets session cookie when curl is not available

runa:
When curl was not available, the session cookie was not being set; so chart/preview was returning 403


gka:
thanks for the fix. merging..

PR #236, opened 2019-02-06
Improve linting and auto-formatting using ESLint and Prettier

sjockers:
This integrates Prettier with ESLint to make sure both can be used side-by-side without getting in the way of each other. It also adds support for linting and auto-formatting Svelte components, including markup.


sjockers:
To be discussed: 

1. Do we really want to automatically format that amount of legacy code? – an alternative would be to remove d4482c8 and instead format and lint pieces of code whenever we touch them.
2. I created an alternative setup (see #237) based on [Healthier](https://www.npmjs.com/package/healthier) which works great for me, has less config and pretty good editor support, despite the fact that the project is a one man show. Which one should we use?

sjockers:
Closing this in favour of #237.

PR #237, opened 2019-02-06
Improve linting and auto-formatting using Healthier and Prettier

sjockers:
[This is an alternative to #236, with simpler config and dedicated extensions for both sublime and vscode]

This integrates Prettier with [Helthier](https://github.com/KidkArolis/healthier) (which is based on the [Standard linter](https://standardjs.com/)) to make sure automatic code formatting and linting can be used side-by-side without getting in the way of each other. It also lints and auto-formats Svelte components, including markup.

Using this setup, you can run the linter via `npm run lint` and the formatter via `npm run format`. The setup also enforces that any files you change comply with this code style using a post-commit hook.  

I'd also suggest you integrate this with the editor of your choice (e.g. [with Sublime](https://github.com/KidkArolis/healthier/blob/master/docs/01-sublime-text.md) or [with vscode](https://github.com/KidkArolis/healthier/blob/master/docs/02-vscode.md)) so that you get immediate feedback to the code you write.


sjockers:
Beware: As our code base does not comply with the linting and formatting rules, this will raise a lot of errors and warnings when you start to use it. Once these are fixed, the setup makes sure that our code style stays consistent.

package.json
```diff
@@ -6,10 +6,18 @@
     "url": "https://github.com/datawrapper/datawrapper.git"
   },
   "scripts": {
+    "format": "prettier --write 'src/**/*.{js,html}'",
+    "lint": "healthier 'src/**/*.{js,html}' && prettier --check 'src/**/*.{js,html}'",
```
sto3psl: Is `healthier` not running the `prettier` check? And maybe the format script could be `healthier --write 'src/**/*.{js,html}'`. I don't know if that's how it works 😅

PR #238, opened 2019-02-07
feat: improve chart rendering by waiting for font load events

sto3psl:
This change observes all fonts that a chart has to load and rerenders the chart after all fonts have been loaded.

Charts in general rely on a lot of absolute positioning and this needs text measurement. To correctly calculate positions fonts have to be loaded, otherwise render code might measure a fallback font that has different letter spacing, letter widths or kerning. This causes text to get cut off when the font is finally loaded.

Currently there is a short flash for rerendering since the chart first renders with a fallback font and then with the custom font. This produces some jumpyness. I think it's fine since the user will see content really quickly. It might also be confusing because of the jumping. So it's up to us to decide if we want to do a first render when the fonts are loaded or do 2 renders and have the flash of unstyled text.

### 1. Flash of unstyled content
fouc

Pro: The first content is visible for the user quickly.
Negative: Chart is flashing and jumping around.

---

### 2. Render after fonts loaded
font-promise

Pro: Chart renders without flashing
Negative: It takes longer to see the first visible content.

---

## About `FontFaceObserver`

[`FontFaceObserver`](https://github.com/bramstein/fontfaceobserver) is a nifty little tool that allows us to listen for font loads. In browsers that support the API it uses `document.fonts`, otherwise it uses scroll and resize events to not block the main thread too much when detecting font load events.

---

## Misc

We do load `Promise` and `Set` polyfills in IE, I checked already :)


gka:
I'd vote for the flash of unstyled content. It just "feels" faster and it's immediately clear that the second rendering is caused by the web font. one question though: In the GIF there's a moment when the chart is rendered but _no text is shown at all_. is this just for demonstration purposes or will there actually be no text shown? I thought there's a fallback font until the webfont is loaded?

sto3psl:
I agree, and prefer to have content as fast as possible.

> In the GIF there's a moment when the chart is rendered but _no text is shown at all_.

I emulated this with slow network speeds to show the difference more clearly. Apparently the canvas rendering is way faster than the other stuff. That's why you see the dots first. On a fast network they appear at the same time. This happens in the current implementation on app.datawrapper.de as well.

PR #240, opened 2019-03-06
Replace environment variables in config.yaml after parsing

gka:
I moved the `parse_config()` up, don't think there's any negative side effect. but maybe you see something that I don't?


davidkokkelink:
no, looks fine for me. appears that it was there just for historical reasons: https://github.com/datawrapper/datawrapper/blob/462141dd22f5213b4b22849320032c25602dc96c/lib/bootstrap.php#L50

PR #241, opened 2019-03-11
collapse folder arrow UI

elanals:
Off the back of @gka's super collapsable folders css hackery, a suggested improvement to the UI for this, inspired by the way sublime does it. 

sublime collapsing folders
Folder collapse UI

### DIFF
--- a/assets/styles/datawrapper/mycharts.less
+++ b/assets/styles/datawrapper/mycharts.less
@@ -199,6 +199,9 @@ body.dw-mycharts {
     .gallery .subfolders .thumbnail .fa {
         color: #999;
     }
+    div.folders {
+        padding-left: 18px;
+    }
     div.folders > ul {
         margin-bottom: 40px;
     }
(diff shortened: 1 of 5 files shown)


gka:
from your GIF it looks as if you have perfectly aligned the triangle and the folder icon:
image

however, it appears that in Firefox the alignment is off by a little bit. could also be that the unicode character depends on local system fonts that are different in Linux than in OSX? I guess also different on Windows, so this might be worth checking as well.

image

To find out if it's just a Linux thing I tried the UI in Chromium... and found another issue. Here it looks as if the collapsed folder triangle is bigger than the open folder triangle:

image

davidkokkelink:
isn't this a bunch of stuff that can be tabled for the svelte rewrite? debugging icon alignment on various operating systems leaves the "20% of the work that produces 80% of the results" territory

gka:
alright I fixed it by using icon font icons instead of platform-dependent unicode characters

PR #243, opened 2019-03-12
Chore: Add unit test setup (AVA)

sjockers:
This adds a setup for unit tests based on [AVA](https://github.com/avajs/ava) and [browser-env](https://github.com/lukechilds/browser-env) for client-side unit testing. 

Use `npm test` to run all tests.

### Running test during development

To run tests during development, you can use `npm run test:watch`, or AVA directly. Using `ava --watch` (or `-w`) will run tests in [watch mode](https://github.com/avajs/ava/blob/master/docs/recipes/watch-mode.md). For more info on running tests, see the [AVA CLI documentation](https://github.com/avajs/ava/blob/master/docs/05-command-line.md).

```
# Run and watch tests:
npm run test:watch # (equivalent to running ava -w)

# You can also run specific test files individually:
npx ava -w ./src/example/ExampleComponent.test.js

# or use patterns:
npx ava -w ./src/example/*
```

### Adding new tests

To add new tests, simply add a file with the test code right next to the file that contains the code you want to test, using the following naming scheme:

```
example
  ├── exampleUtility.js
  ├── exampleUtility.test.js
  ├── ExampleComponent.html
  └── ExampleComponent.test.js

```

To declare a test you use the `test` function you imported from AVA and provide a title and implementation function. The function will be called when your test is run.

```
import test from 'ava';

test('my passing test', t => {
	t.pass();
});
```


test/helpers/setup-svelte-compiler.js
```diff
@@ -0,0 +1,28 @@
+const path = require('path');
+const addHook = require('pirates').addHook;
+const { compile } = require('svelte');
+
+const SVELTE_EXTENSION = '.html';
+
+// We use LESS which requires async preprocessing, which we can't easily do here.
+// To circumvent preprocessing, we simply remove all style tags from the raw code.
+const removeStyles = code => code.replace(/<style([^]*?)>([^]*?)<\/style>/i, '');
```
sto3psl: Nice comment! If you copied the regex from somewhere, it would be cool to have a link in here too. It's fine as is and not too complex anyway.

yarn.lock
```diff
@@ -2,6 +2,358 @@
 # yarn lockfile v1
```
sto3psl: I think we should have either a `package-lock.json` or `yarn.lock`. I think they store `node_modules` slightly differently and if someone adds a dependency with `npm` the next `yarn` user won't have it in their `yarn.lock`.

Any thoughts @gka?

package.json
```diff
@@ -51,6 +57,7 @@
         "rollup-plugin-node-resolve": "*",
         "rollup-plugin-svelte": "4.1.0",
         "rollup-plugin-uglify": "*",
+        "source-map": "^0.5.6",
```
sto3psl: I think you don't use this anymore :)

.gitignore
```diff
@@ -48,6 +48,8 @@ scripts/history.php
 scripts/lib/execute_jobs.lock
 templates/plugins/*
 test/result/*
+test/build.js
+test/build.js.map
```
sto3psl: I think you don't build anymore :)

package.json
```diff
@@ -6,8 +6,10 @@
         "lint": "prettier --check 'src/**/*.{js,html}' && healthier 'src/**/*.{js,html}'",
         "install": "grunt",
         "watch": "watch 'make assets' assets/",
-        "pretest": "cd test/build && rollup -c",
-        "test": "./node_modules/mocha/bin/mocha --require reify"
+        "pretest-legacy": "cd test_legacy/build && rollup -c",
+        "test-legacy": "./node_modules/mocha/bin/mocha --require reify",
+        "test": "ava",
+        "test:watch": "ava --watch --verbose"
```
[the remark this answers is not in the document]
sjockers: Yeah, that makes sense. Thanks for clarifying. I suggest leaving it in anyways, as it makes it clearer what to do for people who are less familiar with AVA.

src/controls/Checkbox.test.js
```diff
@@ -0,0 +1,86 @@
...
+    document.body.appendChild(t.context);
+});
+
+test('Render label with provided text', t => {
+    new Checkbox({
+        target: t.context,
+        data: { label: 'Some text' }
+    });
+
+    t.is(t.context.querySelector('label').textContent.trim(), 'Some text');
```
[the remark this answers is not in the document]
sjockers: Unlike `innerText`, `textContent` does not just render the visible text. Instead, it has all the whitespace that is present in the document, including indentation etc. That's why I have to trim it when comparing, but I agree that it looks weird. 

What do you think about using this instead: 
```js
t.false(t.context.querySelector('label').textContent.contains('Some text'))
```

yarn.lock
sjockers: I second we should remove yarn.lock. Out of habit, I sometimes accidentally use `yarn`, but we should really stick to one package manager.

src/controls/Checkbox.test.js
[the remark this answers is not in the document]
sto3psl: I don't mind the use of `trim()` at all. It accomplishes the same as `textContent.contains` in this case.

I just saw that `Checkbox` adds a `&nbsp;` in front of `label`. That is why I thought you need to trim. Didn't know about the `textContent`, `innerText` differences.

test/README.md
```diff
@@ -0,0 +1,43 @@
...
+# Run and watch test with verbose output:
+npm run test:watch # (equivalent to running ava --watch --verbose)
+
+# You can also run specific test files individually:
+npx ava -w ./src/example/ExampleComponent.test.js
+
+# or use patterns:
+npx ava -w ./src/example/*
+```
+
```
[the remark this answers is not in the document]
gka: @sjockers  I'm just going to put this in and merge the PR for now. feel free to change this if you want

sto3psl:
The PR still had open change requests. I guess the `yarn.lock` can be removed as follow up.

PR #246, opened 2019-03-19
chore/svelte2

gka:
it's time to upgrade the svelte version to 2 (instead of just using the v2 compiler)


gka:
this branch still needs to be properly tested (including full integration tests on staging + with other vis plugins that depend on the svelte controls)

PR #248, opened 2019-05-14
Pass the API domain to svelte apps that define a `api.init` function

sjockers:
This allows you define a simple API wrapper ([such as this one](https://github.com/datawrapper/plugin-admin-users/blob/feature/svelte-user-admin/src/api.js)) to encapsulate the data layer of your svelte app.


gka:
looks good to me

PR #249, opened 2019-05-15
fixed publish component for svelte 2

hanshack:
The Publish.html component no longer uses a **computed** prop to adapt the **embed code** but instead watches the changes with **observeDeep**. Works fine locally but needs testing on staging.


ilokhov:
I tested this locally and it works without errors. But before testing on staging let's finish all the Locator Map issues first - so we can test it all at once.

templates/chart/embed.js
```diff
@@ -1 +1,16 @@
...
+                var iframe = document.getElementById('datawrapper-chart-' + chartId) || document.querySelector("iframe[src*='" + chartId + "']");
+
+                if (!iframe) {
+                    continue;
+                }
+                iframe.style.height = event.data['datawrapper-height'][chartId] + 'px';
+            }
+        }
+    });
+})();
```
gka: this file needs to get simplified again before being pushed to `master`. maybe it got formatted automatically by accident?

ilokhov: Will make sure this is minified before we merge back into the master branch.

PR #253, opened 2019-08-27
Small changes to footer markup + byline append theme option

elanals:
Wrap footer captions in spans to aid direct styling & add theme option to append the byline with a string, e.g company name

`theme.options.footer.bylineAppend`


sto3psl:
@elanals Can you ping me for these changes too, this code is currently a little duplicated and lives here too -> https://github.com/datawrapper/chart-core/blob/master/lib/Footer.svelte. Sorry for the inconvenience.

elanals:
@sto3psl excellent point! Sorry that I didn't think about this, will do in future. What's the best protocol here? Should I adjust `Footer.svelte` with the corresponding changes and PR that as well?

sto3psl:
> Should I adjust Footer.svelte with the corresponding changes and PR that as well?

That would be amazing. Unfortunately the new setup is currently only running for me and @ilokhov but just make a PR and I check it out locally :)

PR #258, opened 2019-10-06
Use default folder for new charts if set in team settings

davidkokkelink:
Allows team admins to set a specific folder where newly created charts should be placed (like a "Drafts" folder for example)

Implementation: If team settings parameter `default.folder` has been set, the folder exists and belongs to the organization of the currently logged in user, move newly created chart to it.


gka:
smart idea

lib/core/build/classes/datawrapper/ChartQuery.php
```diff
@@ -29,7 +29,17 @@ public function createEmptyChart($user) {
         if ($user->isLoggedIn()) {
             $chart->setAuthorId($user->getId());
             $org = $user->getCurrentOrganization();
-            if (!empty($org)) $chart->setOrganization($org);
+            if (!empty($org)) {
+                $chart->setOrganization($org);
+
+                $settings = $org->getSettings();
+                if (isset($settings["default"]["folder"])) {
```
gka: did you test that this doesn't throw an error if `settings.default` doesn't exist?

davidkokkelink: yes, that's how php's `isset` works: http://sandbox.onlinephpfunctions.com/code/a1137a03e9952099e4f6851cd6661ecbdf81e327

PR #259, opened 2019-10-20
Add "move to folder" chart action

davidkokkelink:
This PR adds a new chart action which lets the user place the chart in a folder through a Google-Drive-inspired dialogue. New folders can also be created from that dialogue. It's powered through the new node API, you'll need to clone the API branch `folders` to get it running.

move-folders-3

 A couple of things we could think about to improve this:

- the styles could probably be adapted to look a little more like Datawrapper
- it could be added as a "Move" options on the My Charts/Team Charts page
- it currently opens as a pop-up dialog, laying over the existing layout. This is something we haven't really done so far (and maybe for good reason). Initially I tried to embed the dialog inline (like the PDF export UI, below the chart options), but I couldn't get it to look not weird. So now it's very close to the Google-Drive-original. That may be fine, but further experimentation might be useful (it's just a couple of lines of CSS in `src/v2/folder/App.html`, so easy to try things out)


davidkokkelink:
API changed are now merged so this can be tested on `api:master`

davidkokkelink:
Thanks for the feedback! I would imagine this to be available for all users&customers. The problem (and something we've heard from users but also during my own use) is losing track of where charts even are. You have a chart, it's somewhere buried in one of multiple teams with folders and subfolders. But it's in your dashboard as part of "recent edits". So you can open it, and use "Move" to either place it in a folder, or also simply check in which folder it currently is. As more people use teams in Datawrapper (and we know that number will soon increase), the problem affects more people.

Added @lisacharlotterost as a reviewer here, both from a UX/UI and a Datawrapper power-user perspective.

ilokhov:
Thanks for the clarifications!

The use case that you described makes sense - I did not consider the problem of having too many charts/folders to look through and "losing" charts. At the same time though I wonder, how often will a chart that is lost be in a user's recent edits. These questions should be addressed when we do a re-write of My Charts/Homepage - we can try to integrate the move function in a smart way there.

Sorry - didn't see that Lisa wasn't included here otherwise I would have included her in the reviewers.

davidkokkelink:
It's definitely one of the situations where real-world scenarios might differ from the situation we have as developers where we are primarily working with throwaway test charts. Looks like there are multiple layers of questions here:
- is this even a problem worth addressing
- if yes, is the "chart actions" space in step 4 the right place to do so
- if yes, is the implementation one that makes sense / is consistent enough with the rest of Datawrapper

gka:
I have definitely lost charts a lot of times. I'm in too many teams and then I duplicate some chart and edit it and then I never find it again. Would absolute like this feature

gka:
I'm not really convinced that this is the best place to have this UI though. Also the popup style is sort of new to Datawrapper so it might be worth thinking a little bit more about it with @lisacharlotterost

davidkokkelink:
> one that also shows where the chart is currently stored, ideally with a link to the folder in the mycharts view.
just fyi: both of these things are currently the case (though only after clicking on "move" once). the dialogue opens in the folder where the chart is currently placed, and a click on the folder title (and the little link-icon next to it) opens the respective folder in the mycharts view.

davidkokkelink:
definitely agree that the pop-up style could be improved or ideally even done away with

lisacharlotterost:
Hi everyone, thanks for this great feature! Looking forward to using it. I agree that this is a problem worth addressing & that step 4 is a good place to do so.

The pop-up look should indeed be improved, but IMO only style-wise (and I haven't thought about that part yet). I'm happy with the general UX of this thing. I don't have strong feelings about "should this be a pop-up or not". It's unobtrusive, so that's nice. 

I do think we should show **where the chart is located right now**, to understand *if* you'd actually need to move the chart. 

I also have some opinions about "where exactly in step 4 should this go". Having "Duplicate" as part of the "Export" section doesn't make so much sense, so my first idea was to just group "Duplicate" and "Move" together: 

Screenshot 2019-10-22 at 12 17 51

But that would add an extra headline to the Publish step, which didn't make me happy. So here's another draft (with help texts & without) that clearly separates [what you want to do with the chart outside of *your* Datwarapper, e.g. embed, export, allow for reuse] from [organizational stuff you can do with your chart, e.g. move and duplicate]:

Screenshot 2019-10-22 at 12 18 13
Screenshot 2019-10-22 at 12 18 06

What do y'all think? Happy to move this discussion to Notion, as a "general redesign of step 4" discussion.

elanals:
I think this feature is a great addition in terms of functionality that will ease the workflows for many people a lot, and solve a pretty critical issue with regards to organizing and relocating charts, particularly for situations where people have large organizations with a lot of users, or complex organizational workflows and structures.

Awesome would be if we also eventually had this functionality on the chart dashboard view, but that's not by any means a blocker to launching it in Step 4 for the time being.

I agree with @lisacharlotterost and was going to make the same suggestion, this new option is a good anlass to make more refined groupings of the options in the publish step, and that grouping **duplicate** and **move to folder** would make a lot of sense. My personal preference however would still be to use icons/buttons/similar for the duplicate & move functionality, as I feel like having them as inline text makes these two very useful and critical features a litte too unassuming, I would also push the group up above the river, in accordance with what I consider to be it's relative importance.

elanals:
I also think that adding some kind of **duplicate too folder...** option to the duplicate workflow would be a useful addition, particularly for organizations that translate and have individual folders for different languages for example. (I say this without having though the UI implementation through)

gka:
one problem I have with the popup window is that we don''t use popups for any other of the buttons. Instead clickng a button like svg export opens the UI for that button underneath the button list.

so if the move action was to be added as button, the UI for this action should be shown at the same place we currently show the PDF/SVG export UI.

I definitely like to see where a chart is currently stored without having to click the button first.

Also I wonder if removing the "Duplicate" button would cause a mild user outrage, as many many people rely on this button every day.

lisacharlotterost:
Like so, @elanals ? 
Screenshot 2019-10-22 at 12 52 11

elanals:
Oh she's fast :D @lisacharlotterost. Yes something like that except probably make the styles of groups (icon & title) consistent. I favour the style used for the top three (share/embed/download) because of the compactness. Although I'm aware that this is a direction we're moving away from, not sure if it's necessary for the design of the UI here to be consistent with the refine step or not, in my opinion that isn't so important.

davidkokkelink:
@gka:
> one problem I have with the popup window is that we don''t use popups for any other of the buttons. Instead clickng a button like svg export opens the UI for that button underneath the button list.
> so if the move action was to be added as button, the UI for this action should be shown at the same place we currently show the PDF/SVG export UI.

changed the styling to be in-line instead of popover:
inline-folder

now it looks like we may have a couple of borders / background shades too many?

elanals:
@davidkokkelink the conversation is continuing [here](https://www.notion.so/datawrapper/Restructure-Publish-step-54845f0364d740b3bd39ba0d06cfa3c6)

gka:
In the long term this function shouldn't be part of the chart editor (likewise the duplicate button). Both are general operations on a chart that are useful regardless of the edit progress. It should be easy to move/duplicate unpublished charts as well. So it makes little sense to force users to click on "Publish" in order to get to move/duplicate.

In our current UI the duplicate button is already in the publish step, so we should keep it there for now. However, I'd like to not add the "move" feature to the same spot. It should probably be accessible in all steps of the chart editor, and one obvious place would be somewhere in the 

```
This chart is in Team > Folder
```
...navbar that we added recently. So for now I don't think we should merge this feature as it is, and instead try to redesign it into the chart folder navbar. It could be a small "move" button which then opens a popup window where users can select and search for a target folder.

davidkokkelink:
this will be considered for the new dashboard, but not in this form

PR #264, opened 2019-11-19
Move publish to /src/v2

ilokhov:
After introducing a separate svelte 2 build setup for new components, it makes sense to migrate the publish components (which were already using a separate svelte 2 setup) to use the same setup.

This PR does the following:
- moves publish files from `/src` to `/src/v2`
- adjusts `rollup.config.js` accordingly
- imports everything from `@datawrapper/controls` and `@datawrapper/shared` instead of from deprecated `controls` and `shared` folders in `/src`
- removes some CSS which was used for help bubbles from `datawrapper.css`. Help bubbles have been updated in a newer version of controls and now appear on the other side - this CSS change removes a "?" character which was visible when displaying help messages

There are no new features introduced, ideally everything works exactly the same as before. Works fine on my local setup but I would appreciate a review before merging.


davidkokkelink:
great PR, very nice to have this simplified. works well in my testing.

PR #265, opened 2019-11-21
add support for date formats DD/MM/YY and MM/DD/YY

elanals:
Small PR, all the pieces were already there. Users often upload 2-digit years in both of these formats, would be great to support them for these two, as we already do for the MMM-YY format for example.

I added what I assume are the correct tests to the coffee script, but I don't know how to run it so I'm not totally sure if it's right.


davidkokkelink:
nice, i think this is bloomberg's default format for stock prices, good one to add

PR #267, opened 2019-11-25
feat: use node API for chart editing

sto3psl:
With the the official release of the Node API, I thought it would be fun to put it under a lot of load immediately 😅.

Chart editing now sends data to the new API. Please thoroughly test this and check if any anomalies happen. ✌️

If you want to test locally, you need to have the Node API setup and running. If you are not on the Docker setup, talk to me. If you are using Docker, talk to @gka or @davidkokkelink.


davidkokkelink:
thanks. FYI: charts that use svelte controls save through this function: https://github.com/datawrapper/shared/blob/master/chart.js#L16-L18

to test it, you'll need to change it there and rebuild the respective charts

sto3psl:
After installing the new shared version, a whole bunch of stuff changed. Looked like some static files where not minified. Is all of that fine? Do I need to remove changes?

sto3psl:
Am I assuming correct, that visualization plugins need to update their `@datawrapper/shared` versions too?

davidkokkelink:
> Am I assuming correct, that visualization plugins need to update their @datawrapper/shared versions too?
yes

sto3psl:
Does anything speak against checking out this branch on staging, would like to test it there?

gka:
nope, testing this on staging is fine!

gka:
I think this has been tested enough now?

gka:
rebased to current `master`

PR #269, opened 2019-11-26
Improve custom fields styles

elanals:
Improve the consistency with existing annotate input options, and fix various problematic css.

Before:

image

After:

image


davidkokkelink:
i don't think we should change the styles from what it set is in `@datawrapper/controls`. the purpose of the global css selector was to fix a missing css rule in the annotate tab, not to semantically override the style decisions made in `controls`. if we want the layout as it is in your PR (label above the textbox), which i agree is more consistent, that needs to be added as a property to the `Text` and `TextArea` controls.

elanals:
[ControlGroup](https://github.com/datawrapper/controls/blob/master/v2/ControlGroup.html) seems to have an inline option, not sure if this can be passed from the <text> control?

ilokhov:
@elanals at the moment all Controls are automatically styled inline and to my knowledge there is currently no way to override this, so this is something that would have to be added.

There is also the https://github.com/datawrapper/controls/blob/master/v2/FormBlock.html Control, which should style all form elements inside it as blocks, but I haven't tested how it behaves with multiple elements or `Text`.

elanals:
So is short term inconsistency favoured over short term unclean implementation? Should I get rid of the forced styling?

ilokhov:
In this particular case it would be better to implement these styles in Controls, and then apply them here. This way we don't need to revert the custom CSS here at a later stage and we could also then start using these styles in other parts of the application too.

ilokhov:
I cleaned up a bit, otherwise the change looks good! Let's merge this after the upcoming team changes, to not create unnecessary merge conflicts.

davidkokkelink:
the settings for custom fields have moved into plugin `team-custom-fields`, so this PR needs to move as well. sorry for the trouble!

ilokhov:
@davidkokkelink I can see that `App.html` and `main.js` are still in this repo in `https://github.com/datawrapper/datawrapper/tree/master/src/v2/fields`, although they have now been moved to `team-custom-fields`. Should I remove them from here?

davidkokkelink:
No, what you're looking at is the code to _render_ the custom fields. Those still remain in core for now. The settings however `src/v2/team-settings` have moved to the plugin

ilokhov:
Sorry, my mistake - that makes sense.

In that case this PR can stay, since it targets both files that now live in `team-custom-fields` and also `src/v2/fields/App.html` which is still in this repo.

I will update this PR.

ilokhov:
@elanals feel free to create a PR in https://github.com/datawrapper/plugin-team-custom-fields to apply changes from this PR that have now moved over to the plugin.

PR #270, opened 2019-11-27
add fields for tracking when and by whom a user was invited to a team

gka:
see also https://github.com/datawrapper/orm/pull/14 for orm schema update
* adds two new fields `user_organization.invited_by` and `user_organization.invited_at`.
* fills in default values for existing team members, assuming..
   * each team member was invited by whoever is owner _now_
   * each team member was invited by the time either the team or the user was created (whichever happened last)


davidkokkelink:
looks good to me

davidkokkelink:
the db migration didn't run through successfully on my instance (it only added `invited_by` but not `invited_at`). i tried running it manually and got this error:
sql-error

gka:
ah my bad. need to allow NULL until the `invited_at` dates have been filled in

gka:
should run through now

davidkokkelink:
yes, works now :+1:

PR #271, opened 2019-11-28
Don't strip tags in byline so that bolding is possible

elanals:
they get purified in chart.twig anyway


davidkokkelink:
@gka the byline still gets server-side purified in `chart.twig` as elana mentioned. this is just removing a second layer of stricter stripping which is inconsistent with the other fields, where our default whitelist applies (defined in `twig-init.php`)

gka:
alright

PR #272, opened 2019-11-28
Team invite ui update

ilokhov:
This PR adds visual feedback to the **Invite team member** section in team pages (`/team/{team_id}/members`).

We now show a success message if someone has been invited or added and an error message in case there was an error. In addition, we no longer show the spinner once the request has finished processing.

Example:

Screenshot 2019-11-28 at 18 34 03


elanals:
With permission I would make the email address, and the type (Team Member) bold in the translations sheet. Thoughts?

ilokhov:
Neat idea @elanals, I just added this!

gka:
It would be nice if we could also show _why_ a user couldn't be invited/added. There are only three scenarios that come to my mind:

* The user could not be invited because they are already a member of the team. (`400 bad request`)
* You have to activate your account before you can invite users to your team. ( `401 unauthorized`)
* You have already invited 20 users who haven't accepted your invitation yet. Please wait until they accept the invitation to invite more users. (`406 not acceptable`)
* The user could not be invited due to a server error. We are looking into this. If you need to invite someone urgently, please contact support@datawrapper.de (`500 internal server error`)

Since only admins can add users (as opposed to inviting them) I don't think we need to care about error messages for this case.

ilokhov:
Just updated the PR to show detailed error messages. We now show custom messages for `400`, `401` and `406` and a generic Server Error message for any other errors. We also include the current limit of max team invites in the message on `406`.

I also included a custom message for `400` for admins when trying to add users, since I thought this would be convenient for admins to know.

This PR depends on: https://github.com/datawrapper/api/pull/69

davidkokkelink:
thanks! i've updated the translation strings a bit. looks all good to me

PR #273, opened 2019-11-28
use JSON-based clone instead of $.extend

gka:
during development of the new `d3-maps` controls I saw that sometimes the $.extend based clone of the chart attributes would not deep-copy all attributes correctly, so that some attributes where changed in the `__dw.old_attrs` directly which prevented re-rendering of the chart (because no change could be detected).

this PR fixes this by using `JSON.parse(JSON.stringify(..))` to store a copy of the previous attributes for change-detection.

see https://github.com/datawrapper/datawrapper/pull/273/commits/b95cd466481c985e11d67eba2820e8e99265eeff for the actual change. the rest is just linting of the file.


davidkokkelink:
wow, nice catch. pretty fundamental to our UI, surprised this didn't cause more problems

PR #278, opened 2019-12-06
Fix number format

davidkokkelink:
This PR fixes two different bugs regarding number formatting:
- When `number-format` is set to `auto`, don't set the `number-divisor` to `0`.
- Use cloned metadata instead of operating on the reference: When number-format is set to `auto`, this part of the code actually sets the number format based on `dw.utils.significantDimensions`. However, this happened on the actual metadata object instead of on a clone. Due to this, when you set it the `number-format` to `auto`, it would simply hardcode the return value on the next call of `columnFormatter`. Because of bug #1, this would also reset the `number-divisor`. On the subsequent call, the `number-format` value would no longer be `auto`, and the hardcoded values would take place again. All in all, totally not what anybody would expect from a function that takes a column and returns a column formatter function.


ilokhov:
Thanks for looking into this!

The bug itself is fixed but I encountered another issue: when setting a value in "Divide/Multiply by" and then selecting a different column, instead of showing the default "(no change)", it shows the value from the previously selected column. Notice how column C shows `100` when I switch to it.

columnFormat

Maybe the logic in the `describe` component also needs to be updated?

davidkokkelink:
Good catch. This was caused by an incomplete default setting that was masked by the other bug. Fixed now

PR #280, opened 2019-12-12
Convert booleans in theme json into fake boolean before passing to less

elanals:
So that real booleans can be used in theme json.


davidkokkelink:
nice - no more `fakeBoolean` in the theme schema (and data)

gka:
it'll also be pretty easy to fix the data once the schema is complete, bc we can use it to identify all the fakeBookean options. but we still need to convert the theme.data column to JSON, it's still stored as longtext.

PR #281, opened 2019-12-12
change theme.data to json column

gka:
changes `theme.data` column type from `longtext` to `json` so we can run db migration queries on theme data in the future. will likely need some more changes in `orm` & `api`.


gka:
merged and deployed

PR #283, opened 2019-12-16
send only chart attributes required for rendering

gka:
this PR fixes a bug that happened bc we were sending the entire svelte controls state to the iframe preview, which included unserializable objects like `dataset` etc. Instead we're now just passing attributes on a need-to-render basis.


gka:
@davidkokkelink any other attribute that we should pass to the iframe?

davidkokkelink:
no, object seems complete to me

PR #285, opened 2019-12-19
Set clone of custom fields object to metadata to ensure that change is registered

elanals:
Also removed unused helper


gka:
makes sense. would you mind removing the unrelated build files from this PR before merging? they don’t belong here and should be added in a separate PR / commit. Referring to all the files like www/static/svelte/controls.js etc

elanals:
@gka Sorry for all the back and forth commits when sorting this out. Is this fine now? Not sure if any of the other build files were needed, I think the changes happened in the other build files because i npm i'd

gka:
looks good! just squash and merge and the back and forth commits will be gone forever :)

PR #286, opened 2020-01-06
pass chart to publish_files hook handler

gka:
To support multiple team-dependent publish providers the publishFiles hook needs the `$chart` object (so it can decide where to publish based on the team settings etc).


gka:
also see https://github.com/datawrapper/plugin-publish-s3/compare/accept-chart-first-param?expand=1

PR #288, opened 2020-01-08
use asset_domain defined by hook if available

gka:
* moved underscore and jquery to our asset S3 bucket
* allowed overwriting the core `asset_domain` by plugins like `publish-cloud`
* we're now appending the `/assets` path to the "domain", so this update will require removing the trailing `/assets` from the `config.yaml` setting as it is from now on appended automatically


davidkokkelink:
works!

PR #290, opened 2020-01-14
Catch font loading errors

davidkokkelink:
To get rid of these types of errors that plague our users' consoles:
timeout


davidkokkelink:
linter hates me for this because i don't "handle the error object". maybe @sto3psl has some insight on the best way to fix this?

sto3psl:
@davidkokkelink hmm I think you could print something with `console.warn`.

Maybe `console.warn('Font observer timed out.')` or something. Otherwise an `//eslint-disable-line` can help too 😄

davidkokkelink:
i don't want to log, so i went with `// eslint-disable-next-line`

davidkokkelink:
you're right, changed it

PR #292, opened 2020-01-17
rewrite signin modal in svelte

gka:
work in progress.

goals of the refactoring:
* improve error handling of signup modal
* make it easier to switch to v3 API signup endpoint (not sure if we want that already)


sto3psl:
Nice! Some thoughts about the v3 signup endpoint.

We should be able to switch without issues but someone else than me should properly check if all requirements and the correct functionality is implemented. It's has been a long time since I wrote that. We don't have the password hash migration enabled anywhere but if we want to do that in the future, this should get checked out too. I wrote a bunch of tests for it so it should be good but better safe than sorry. When the time comes I'm also happy why I implemented that and what it does.

gka:
This PR now includes a significant change in our existing PHP API: the endpoints `POST /auth/login` and `POST /users` will now return proper HTTP error codes if the request fails. Need to think carefully about side-effects, especially for other components that use these endpoints.

gka:
also this PR depends on another PR https://github.com/datawrapper/shared/pull/23

PR #293, opened 2020-01-23
Use v3 API routes for signin

gka:
depends on https://github.com/datawrapper/api/pull/109 to be merged 

## PHP-API changes
* **removed legacy api endpoints** because they are incompatible with new bcrypt hashes
   - `POST /auth/login`
   - `POST /users`
   - `POST /auth/logout`
   - `DELETE /users/:id`
   - `PUT /users/:id` 

## Build setup changes
* Svelte components in `src/v2` are now being transpiled to ES5 **using Babel** instead of **buble** to ensure IE11 compatibility 

## Frontend changes
* the sign-in and invite forms now **uses the v3 api** for login and sign-up. the old twig code has been replaced by the following Svelte UI components:
   - **signin**: [src/v2/signin/SignIn.html](https://github.com/datawrapper/datawrapper/blob/signin-use-v3-api/src/v2/signin/SignIn.html)
   - **invite**: [src/v2/invite/Invite.html](https://github.com/datawrapper/datawrapper/blob/signin-use-v3-api/src/v2/invite/Invite.html)
* the **account settings now use v3 api** instead of legacy api for changing password, email and deleting account (which again, are now incompatible with new bcrypt hashes)
* removed obsolete signin form in **core publish step for guests** in svelte visualizations (e.g. locator maps). The code was in `src/signin/SignInForm.html`, and it was only used in `publish/sidebar` app. Instead we're now just showing a link that opens the signin modal.
* password requirements have been syncronized across Datawrapper: we now require a **minimum length of 8 characters**
* we're now loading [`zxcvbn`](https://github.com/dropbox/zxcvbn) to give the user feedback about their password quality. This is feedback is only informative and does not prevent users from picking a certain password (yet). Since the `zxcvbn` library is rather big (> 700kb) we are lazy-loading it (via requirejs) once a user starts typing a new password into the sign-up form.
* the new **invite** Svelte component is now being used for the "send embed code via email" signup, the "invite user to team" and the "reset password" process
   - grafik
   - grafik
   - grafik


sto3psl:
A small thing I just noticed. The contrast ratio (2.67) for the information text here is pretty low. In combination with the small font and a non high DPI display I had some trouble reading it.
image

gka:
alright, I increased the font-size by 1 pixel. But the colors come from [Bootstrap](https://getbootstrap.com/2.3.2/base-css.html#typography) so I'd probably hold off from changing it until we're doing a more fundamental redesign 
grafik

sto3psl:
Yes, I realized that after doing a quick search through our code base. It's interesting to me that the contrast in the current Bootstrap version is even worse. It works fine as background color, not so much as text color.

PR #298, opened 2020-01-31
this needs to be done as a migration of the subscriptions plugin

EX0l0N:
If we set up the database solely from the migrations folder, than we can not change this table here, because it does not exist.

This table is created by the subscriptions plugin.

I believe we should set up the database by first running all migrations of the base datawrapper directory and then run the migrations of every plugin one by one. There shouldn't be any order of the plugins required.

Leaving it the way it is now, we'd need to install some of the migrations from datawrapper base, then install subscriptions and then again run the rest of the DW migrations.

I will write another PR for the subscriptions plugin.


davidkokkelink:
Makes sense :+1: 

> I believe we should set up the database by first running all migrations of the base datawrapper directory and then run the migrations of every plugin one by one. There shouldn't be any order of the plugins required.

Yes - the way we execute migrations is by running `php scripts/sync-db.php`, which will take care of it. Please don't execute migrations manually

Some meta-feedback around the PR, not the code: A title like `Remove chargebee-invoices migration from core migrations` would be both shorter and more informative (just for the future)

davidkokkelink:
I think the table `chart_access_token` is also not part of core, but created by `plugin-onetime-access` and the migration should move there. Same goes for `river_chart_tag`

PR #302, opened 2020-02-12
replace buble with babel to get async/await transforms

gka:
long story short: buble doesn't support async/await transformations, a feature we need to get our code to run in IE11. babel does support async/await transformations via plugin. 

currently deployed on staging if anyone wants to test in IE11 (it works)


gka:
@sto3psl requested review because I just copied the babel config from chart-core. maybe you can optimize the config before I merge this?

sto3psl:
Checking this on IE11 currently and I get quite some problems. I'm not entirely sure where all this code runs so what I'm about to write might be unrelated to this PR.

1. Locator maps don't load at all.
```
SCRIPT1006: Expected ')'
controls.js (1,924687)
```
The line looks like this:
```js
geocode:_debounce(async function(e){var t=this;this.set({searching:!0});
```
IE doesn't understand `async`. Since this is from the Locator maps controls I assume it's a problem there.

---
```
SCRIPT1003: Expected ':'
visualize (722,32)
```
```js
chart.on('state', ({changed, current, previous}) => {
```
IE doesn't understand destructuring.

2. Pie charts "Refine" tab doesn't render (endless loading spinner) and the visualization renders no chart.

```
Uncaught (in promise) TypeError: Unable to get property 'contentWindow' of undefined or null reference
   at getContext (https://app.staging.datawrapper.de/chart/glwMe/visualize:1507:9)
   at Anonymous function (https://app.staging.datawrapper.de/chart/glwMe/visualize:1530:9)
   at g (https://app.staging.datawrapper.de/polyfill:31:2556)
   at f (https://app.staging.datawrapper.de/polyfill:31:1055)
   at Anonymous function (https://app.staging.datawrapper.de/polyfill:31:4947)
   at e (https://app.staging.datawrapper.de/polyfill:31:1189)
```

```
SCRIPT438: Object doesn't support property or method 'assign'
preview (1517,5632)
```
I assume something is off with the polyfill loading. Looking at the network tab. The polyfill request returns `404`. Do we even host the polyfills on `https://datawrapper-stg.dwcdn.net`?
image

---

Otherwise the app seemed to work fine for me. Login worked and general navigation too. I was a little worried if there is a `Promise` polyfill. Since there is one all should be fine.

PR #305, opened 2020-02-26
Allow publishing into subdirectory

davidkokkelink:
This PR adds the following functionality:
- Use **relative paths** for JavaScript libraries in published charts to allow publication into subdirectory
- Add core config option `hash_publishing` to publish into a directory name based on the hash
- Remove generation of empty unused CSS file during chart publication
- Remove some unused code regarding to legacy thumbnail generation


davidkokkelink:
Changed hashing behavior to use `config.chart_id_salt` (if configured)

PR #306, opened 2020-02-27
Allow variables in share url tempaltes

davidkokkelink:
We have a hook called `CHART_ADD_SHARE_URL` to allow the definition of additional share URLs to be displayed in the publish UI (for example used [here in plugin chart-display](https://github.com/datawrapper/plugin-chart-display/blob/master/init.php#L84-L91)). They appear here:
share-url

Currently, they can only use one hardcoded variable `%chart_id%`. This PR extends it so the share URL templates can also use info from the metadata, like `%metadata.custom.cmsID%`.

The use case are CMS integration plugins that might want to link to the chart in some asset store where it is identified not through its Datawrapper ID but through some other identifier (for example set earlier as a custom field)

The second change in this PR is that the publish implementation used in locator maps sends out the same `postMessage` after a chart publish as all other chart types.


src/v2/publish/App.html
```diff
@@ -337,16 +350,14 @@ <h2>{@html __('publish / share-embed') }</h2>
...
-                if (window.parent) {
-                    window.parent.postMessage(
-                        {
-                            source: 'datawrapper',
-                            type: 'chart-publish',
-                            chartId: chartInfo.id
-                        },
-                        '*'
-                    );
-                }
```
gka: I don't see how removing this safety check will improve the situation. If there is no `window.parent` then `window.parent.postMessage` is undefined

sto3psl: `window.parent` is always truthy. That makes the check obsolete.

> The Window.parent property is a reference to the parent of the current window or subframe.
>
> If a window does not have a parent, its parent property is a reference to itself.
> https://developer.mozilla.org/en-US/docs/Web/API/Window/parent

PR #309, opened 2020-02-29
Simplify publishing logic, use sign up from the core

davidkokkelink:
In the past, we had a pricing structure that prohibited the publishing of charts under certain circumstances (view limits, limits by chart type). Due to this, we implemented a fairly complex publishing logic that allowed plugins to overwrite the publishing behavior. Since then, we have simplified our pricing structure. This PR addresses that. 

## Technical changes
- remove hooks to overwrite publish pages (`PUBLISH_CUSTOM_PAGE_GUEST, PUBLISH_CUSTOM_PAGE_PENDING_ACTIVATION, PUBLISH_CUSTOM_PAGE_NOT_ALLOWED, PUBLISH_CUSTOM_EXPORT_ACTIONS_HEADER`)
- move the pages displayed for guests and unanactivated users in step 4 from `subscriptions` to the core
- fix a bug in the step 4 sign up flow with the new API (call `/v3/auth/signup` instead of `/v3/users`, as only the former creates the user _and logs them in_)
- have both the svelte-based new chart editor, and the twig-based chart editor use the same svelte components for the sign up (`src/v2/publish/guest/Guest.html` and `src/v2/publish/pending-activation/PendingActivation.html`). this way, the UI is consistent and we can make UX improvements in one place.
- use the new node API for the "resend activation link" button
- remove the then no longer needed PHP endpoints for resending the activation link
- remove the then no longer needed PHP email hooks
- move the river component for the svelte chart editor (`AddToRiver.html`) to the river plugin (previously it was hardcoded into the subscriptions plugin). now, river plugin adds UI through a new hook `SVELTE_PUBLISH_AFTER_EMBED`, very similar to how it works in the existing twig editor.

This is all quite intertwined, which leads to this rather large PR.

## UX
From a **UX perspective**, there will be **no visible changes to users of datawrapper.de**. For people using only the PHP core without the `subscriptions` plugin, two things will look different now:

**1. publish step as a guest:**

before:
guest-ux-old

after:
guest-ux-new

**2. publish step with unactivated account:**

before:
pending-activation-ux-old

after:
pending-activation-ux-new

given that all the necessary api endpoints and controllers to support the step-4-based sign up flow are already part of the core & api, i felt it was a sensible choice to also move the UI for this. looking at the before/after, i don't think anybody would complain. if we think the texts are too salesy for datawrapper core (which i wouldn't agree with, they are fairly generic), we could also add hooks (again) to override them - or alternatively, tone them down a bit. i'd be fine with that, i think it's certainly better than having the complexity of the two override systems in place.


gka:
I still think the "for free" and "get in touch via support@datawrapper.de" shouldn't be part of our core app, so I wouldn't mind if the core would provide a way for the `subscription` plugin to replace the HTML that's being shown in these step.

But I do agree that the process itself should live in core to make our lives easier.

davidkokkelink:
Added hook `PUBLISH_TEXT_GUEST` to replace the marketing text we use for sign up. Core UX now looks like this:

core-ux

UX with `subscriptions` plugin remains unchanged.

PR #316, opened 2020-03-13
create guest_session index for chart table

EX0l0N:


davidkokkelink:
I really would like to have all indexes in migrations

EX0l0N:
That's a really good idea. I will take a dump of an initial DB as seen by docker and my ansible setup and diff that against the create statements of our life server.

EX0l0N:
We do have some more important differences between what we build in docker and what is live.
For example the `chart.metadata` field is `DEFAULT NULL` BUT `NOT NULL` in docker builds.

Furthermore there a a lot of charset differences where some table columns in at prod still have an explicit `latin1` setting.
All in all there are far to many columns that have an explicit setting in both configurations, but I guess that's nothing we can fix anymore, without looking very closely at prod.
Most of the time these are `utf8` for docker and `latin1` at prod, although tables have a table wise default of `utf8m4`.

This is just to raise some attention do the man differences we have.
This PR will only cover adding indexes.

EX0l0N:
I will continue this in a more descriptive branch.

PR #319, opened 2020-03-24
allow override of axis.optional setting using overrideOptionalKey

gka:
the idea is that a visualization can define an "override key" for the `optional` setting of an axis:

```php
$axes['colors'] => [
    'accepts' => ['text'],
    'optional' => true,
    'overrideOptionalKey' => 'visualize.colorByColumn.enabled'
];
```
when I tested this in the old bar charts I had to manually reload after setting the override column as otherwise the remaining controls don't know about the change.


sjockers:
Not sure what's the deal with the failing `npm ci` on CircleCI. It seems to be completely unrelated to this change. 

I've already tried re-running on CircleCI and I can't reproduce the problem locally. Maybe related to private npm packages? Any ideas @gka @davidkokkelink?

sto3psl:
> Not sure what's the deal with the failing npm ci on CircleCI. It seems to be completely unrelated to this change.

The `@datawrapper/polyfills` package was set to `private` on npm which is why locally it worked for you bit CircleCI couldn't find it. That might have been me with a prerelease by accident. 

I changed it back to public and now CI works again. Sorry for the inconvenience.

PR #320, opened 2020-03-25
new node publishing & preview

gka:
opening a Draft PR for this


package.json
```diff
@@ -4,12 +4,9 @@
...
         "format": "prettier 'src/**/*.{js,html}' --write",
         "lint": "prettier --check 'src/**/*.{js,html}' && healthier 'src/**/*.{js,html}'",
-        "install": "grunt",
         "watch": "watch 'make assets' assets/",
         "pretest-legacy": "cd test_legacy/build && rollup -c",
-        "test-legacy": "./node_modules/mocha/bin/mocha --require reify",
-        "test": "ava && cd dw.js && npm install && npm test",
-        "test:watch": "ava --watch --verbose"
+        "test-legacy": "./node_modules/mocha/bin/mocha --require reify"
     },
```
gka, line 10:
```suggestion
        "test": "ava",
        "test:watch": "ava --watch --verbose"
    },
```
I think we want to keep the `ava` tests

www/static/js/dw/backend/popupChart.js
```diff
@@ -1,34 +1,35 @@
-
 define(function() {
-
     return function(chart_id, preview, closeCb) {
-
-        $.getJSON('/api/charts/'+chart_id, function(res) {
-            if (res.status == "ok") {
+        $.getJSON('/api/charts/' + chart_id, function(res) {
```
gka, line 3: why not use the `v3` api here while we're at it?

package.json
sto3psl: Good catch. Didn't know that this repo had `ava` tests. Then we should restore the CI test job too.

sto3psl: They are back: aa004c1

PR #326, opened 2020-04-01
quick copy embed code feature

gka:


elanals:
No screenshot? Am I supposed to check out the branch? :D

elanals:
Also wow that was fast, thank you!

davidkokkelink:
This PR uses ES6 features in non-rolluped code, which would probably break in certain browsers. Also there's a console log that probably isn't needed in production.

gka:
@davidkokkelink you're talking about the two `const` statements, right? I'll replace them with `var`. the rest should run in IE9

gka:
I deployed it on staging

davidkokkelink:
Thanks! Should be good now

gka:
tested it on IE11 and it worked. can''t test on IE10 because login is broken

davidkokkelink:
Great - I noticed there's a translation missing for "copied to clipboard"

gka:
none of the 3 buttons is translated on production.

gka:
I am closing this for now. We can re-open the PR once we're done with the node publishing so you can properly review this.

PR #328, opened 2020-04-01
Don't rewrite metadata for print charts

davidkokkelink:
In the past, we had a 'magic' rewrite of metadata settings for customers using a print mode provided by plugin `webtoprint`. This approach had various problems and was not consistently supported throughout the application. This PR removes it.


davidkokkelink:
Instead, the `webtoprint` plugin UI switches between two separate charts. A metadata-level object `custom.webToPrint` is used to maintain the connection between the two.

PR #333, opened 2020-04-07
Fix issue with double-quoted empty values in dataset

davidkokkelink:
See [this discussion](https://www.notion.so/datawrapper/Undefined-vs-Null-CSV-parsing-of-datasets-with-empty-cells-denoted-with-empty-quotation-marks-37c01c690891401cb04a0c19af6fe58b?d=d92a8321-122c-4ea1-bd9b-f58c473656d4)


gka:
it would be nice to add tests for this behavior. the way the current tests work is to go to do

```bash
cd services/datawrapper/dw.js
npm test
```

gka:
this would be the existing test to extend: https://github.com/datawrapper/datawrapper/blob/master/dw.js/test/datasource-delimited-test.coffee#L1

gka:
I'm going to look into why the existing test fails

gka:
alright, I added a test and it appears that the empty quotes `""` are still parsed as `'undefined'`

elanals:
I think this PR was just meant to fix the fact that the CSV parser breaks entirely when values in the last column are undefined, due to the `.length` measurements on these. I don't think David intended to fix the parsing itself. Maybe misunderstanding something here.

gka:
I fixed it!

gka:
btw I also pushed this fix to `shared`: https://github.com/datawrapper/shared/commit/202a664b9edcf2c0ab06c60d1305b32fcec29d60

PR #335, opened 2020-04-16
remove dw.js build and use it from frontend service

sto3psl:
Removing `dw.js` build in reference to https://github.com/datawrapper/chart-core/pull/17


templates/core.twig
```diff
@@ -74,7 +74,7 @@
     </script>
     {% endif %}
     {# javascript core 2.0 #}
-    <script type="text/javascript" charset="utf-8" src="{{ config.static_path }}/js/dw-2.0.js?v={{ COMMIT_SHA }}"></script>
+    <script type="text/javascript" charset="utf-8" src="//{{ DW_DOMAIN }}/core/dw-2.0.min.js?v={{ COMMIT_SHA }}"></script>
```
gka, line 77:
```suggestion
    <script type="text/javascript" charset="utf-8" src="//{{ DW_DOMAIN }}/lib/chart-core/dw-2.0.min.js?v={{ COMMIT_SHA }}"></script>
```

PR #336, opened 2020-04-16
adjustments for dw.js refactoring

gka:
these are the minor changes required to get the [new dw.js](https://github.com/datawrapper/chart-core/pull/18) working.

* chart.load now returns Promise instead of $.Deferred


elanals:
can we merge this so that I don't have to be in this branch for things to work?

PR #337, opened 2020-04-20
set SameSite attribute in Set-Cookie Header

sto3psl:
Chrome and Firefox will "soon" change the default value for the `Set-Cookie` attribute `SameSite`.
* https://www.chromium.org/updates/same-site
* https://blog.chromium.org/2019/10/developers-get-ready-for-new.html
* https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie/SameSite

*Chrome 80 was supposed to introduce the change but they reverted it because of the whole pandemic issue.*

They will switch to `SameSite=Lax` by default when the attribute is not set. This doesn't affect us directly (our app will still work) but we get some warnings in DevTools.

> Cookie “DW-SESSION” will be soon rejected because it has the “sameSite” attribute set to “none” or an invalid value, without the “secure” attribute. To learn more about the “sameSite“ attribute, read https://developer.mozilla.org/docs/Web/HTTP/Cookies

When enabled in `about:config` (Firefox):
> Cookie “DW-SESSION” has “sameSite” policy set to “lax” because it is missing a “sameSite” attribute, and “sameSite=lax” is the default value for this attribute.

By adding `SameSite=Lax` to our `Set-Cookie` header, we get rid of the warnings, which can trigger quite often.

Screenshot 2020-04-20 at 17 26 34

I made the according changes to `frontend` and `api` too:
* https://github.com/datawrapper/frontend/tree/samesite
* https://github.com/datawrapper/api/tree/samesite

This is not super urgent but since it's a quick fix I thought I open a PR.


davidkokkelink:
Yeah... The issue that it actually _does_ impact us since customers use Datawrapper integrated into their CMS (embedded as iFrames) which requires `SameSite=None` to work. So this browser update risks existing integrations. I wanted to investigate first whether there are other solutions to getting third-party cookies to load the way we need them to before moving everybody to `Lax` and breaking this (I do realize that with newest Chrome there are already issues)

davidkokkelink:
So unfortunately we can't merge this for now

sto3psl:
Thank you for the feedback. I kind of figured that on-premise might be another story.

According to MDN we can still use `SameSite=none` but it requires the `Secure` attribute, which itself requires HTTPS. I hope all the on-premise instances are using HTTPS.

My suggestion would be to make this configurable. Use `SameSite=Lax` by default, `Secure` always (except dev setups) and `SameSite=none` on the instances that need it.

API and Frontend already send the cookie as `Secure` when HTTPS is used. I would have added it to the PR here but I'm not sure how I can detect it in PHP and if there is a config variable or another way.

davidkokkelink:
This is not about on-premise instances, it's about customers using our hosted solution who use the [login token API](https://developer.datawrapper.de/docs/getting-started-with-integrations) to generate a token-based login URL server-side, then embed those as an iFrame in their CMS. This breaks with `SameSite=Lax` , as the iFrame embedded on a third-party domain won't get the cookie set

gka:
Let's set this PR back to draft state and find a solution after we launched the new publishing & preview, alright?

davidkokkelink:
Closing this as we have decided to go with a different approach (documented in Notion)

PR #338, opened 2020-04-30
remove obsolete JS, import from shared & controls + update build system

gka:
quite a few changes, but I really only went through and replaced 
```js
import clone from '../shared/clone.js'
```
with 
```js
import clone from '@datawrapper/shared/clone'
```
etc. we can also merge this post-launch!


ilokhov:
Just tested this quickly locally and everything seems to work. But I would much rather prefer merging this after launch.

> but I really only went through and replaced...

Doesn't quite look like that to me 😅

gka:
the merging sort of destroyed this PR, so I opened a new one here https://github.com/datawrapper/datawrapper/pull/339

PR #341, opened 2020-05-05
Delete unused files

davidkokkelink:
This PR deletes a bunch of unused PHP, twig and JavaScript files.


controller/plugin-templates.php
```diff
@@ -1,24 +1,5 @@
 <?php
```
gka, line 1: did you check that no single plugin is using these?

www/static/js/dw/chart/publish.js
```diff
@@ -1,112 +0,0 @@
...
-            var w = $('#embed-width').val(), h = $('#embed-height').val(),
-                embedCode = $('#iframe-code').html();
-            embedCode = embedCode.replace(/width="\d+"/, 'width="' + w + '"');
-            embedCode = embedCode.replace(/height="\d+"/, 'height="' + h + '"');
-            $('#iframe-code').html(embedCode);
-
-            resizeIFrame();
-        });
-
-        $('.chart-actions .action-duplicate a').click(triggerDuplicate);
```
gka, line 20: if you delete this, the chart duplication UI will stop working

controller/plugin-templates.php
davidkokkelink: yes

www/static/js/dw/chart/publish.js
davidkokkelink: thanks, fixed

davidkokkelink:
Thank you for the review, I have fixed the chart duplication bug. I did code searches for all the functionality that I removed, and tested locally to ensure that nothing breaks.

I know this is an annoying PR to review, but I think it's important that we do it:
- duplicate and dead code is confusing for new developers
- having effectively unmaintained code be executed server-side on our production systems is a security risk
- reducing the scope of this repository brings us closer to our goal of getting rid of PHP entirely

I have deployed this branch on staging for further testing.

davidkokkelink:
This was running on staging for close to two weeks without problems now. Planning to merge this tomorrow unless any issues are raised.

davidkokkelink:
and off to production it went

PR #343, opened 2020-05-09
Use v3 API to store and read chart assets

davidkokkelink:
This PR changes the `Chart->writeAsset` and `Chart->loadAsset` functions to call the v3 api instead of storing assets to S3 or the local filesystem itself. 

### Advantages
- Simpler configuration: only the v3 API is responsible to decide where and how to store chart assets, there's no need to keep two configuration files in sync.
- Security: The PHP instance does not need to have access to the private S3 buckets anymore, so we need less AWS roles and keys.

### Disadvantages
- Performance: The PHP app needs to call the v3 API first, which calls the S3 API, instead of calling S3 directly. This should have **slightly worse performance for endpoints that call `loadAsset` or `writeAsset`, though it wasn't noticeable to me on staging**, where it is currently deployed.

### Implementation choices
- I decided to add a private instance variable `$chart->assets` to the `Chart` class to cache assets, so if `$chart->loadAsset` is called multiple times over the course of a request, it is loaded only once from the remote API.
- I also changed the behavior of the asset functions to not return the file path where the asset was stored, as it leaked (somewhat) private information like the name of our S3 bucket.


lib/core/build/classes/datawrapper/Chart.php
```diff
@@ -144,70 +146,46 @@ protected function getDataFilename() {
...
+     * writes any asset to the asset store using v3 api
      */
     public function writeAsset($filename, $data) {
-        $cfg = $GLOBALS['dw_config'];
+        if (!$this->assets) $this->assets = [];
 
-        if (isset($cfg['charts-s3'])
-          && isset($cfg['charts-s3']['write'])
-          && $cfg['charts-s3']['write'] == true) {
+        [$status, $body] = call_v3_api('PUT',
```
gka, line 154: I think we should check the `$status` of the v3 api request and throw some sort of error if the request failed.

lib/core/build/classes/datawrapper/Chart.php
```diff
@@ -144,70 +146,46 @@ protected function getDataFilename() {
...
-            }
-        } else {
-            $filenamePath = $this->getDataPath() . '/' . $filename;
-            if (!file_exists($filenamePath)) {
-                return '';
+            if ($status == "404") {
+                $this->assets[$filename] = "";
             } else {
-                return file_get_contents($filenamePath);
+                $this->assets[$filename] = gettype($body) == "string" ? $body : json_encode($body, 1);
```
gka, line 184: same here, I think we should check for a `200` success code to catch other server errors. otherwise this code could try to json encode error messages etc.

gka: also, why aren't we using `json_encode_safe` here?

lib/api/charts.php
```diff
@@ -162,9 +162,9 @@ function if_chart_exists($id, $callback) {
         }
         $data = $app->request()->getBody();
         try {
-            $filename = $chart->writeData($data);
+            $chart->writeData($data);
             $chart->save();
-            ok($filename);
+            ok();
```
gka, line 167: this changes the existing API signature, but I'm not sure we ever used the returned filename for anything...

davidkokkelink: yes, this was intentional

lib/core/build/classes/datawrapper/Chart.php
davidkokkelink: yes, makes sense

lib/core/build/classes/datawrapper/Chart.php
davidkokkelink: didn't know / forgot about `json_encode_safe`, makes sense though

davidkokkelink:
Thank you for the feedback! I've added proper error handling and used `json_encode_safe` instead of `json_encode`

PR #344, opened 2020-05-12
Migrate describe step to Svelte v2

gka:
While I was at it I also updated Handsontable to 5.0.2 (the latest version we're able to use since they started to drop free features in v6 before killing the MIT license entirely in v7).

This PR also implements proper display of parsing errors for our new `expr-eval` based computed columns, as well as fixing auto-completion for new functions.

Peek 2020-05-12 14-11


gka:
Lisa found one "last" bug: links don't show up in the data table anymore

gka:
interesting, thought we're already using babel

gka:
which file?

davidkokkelink:
`describe.js` - it's a bit strange

gka:
no more `=>` in `describe.js`!

PR #345, opened 2020-05-18
Enable Svelte annotation controls

ilokhov:
This PR includes a new hook `AFTER_CUSTOM_ANNOTATION_CONTROLS` and template `svelte-annotate.twig`, which can be used by visualizations to render Svelte annotation controls.

To render annotation controls for a specific visualization, `svelte-annotations` has to be set to `true` and `AFTER_CUSTOM_ANNOTATION_CONTROLS` needs to hooked into in `init.php`. [See how that looks in d3-lines here](https://github.com/datawrapper/plugin-d3-lines/blob/f9d1c13849f51c9e6e09c9de490918cd94d25215/init.php).

I would like to have used an existing hook `CUSTOM_ANNOTATION_CONTROLS`, however as far as I know it's not possible to control the render order within a hook. So custom fields, which use the same hook, would currently render after annotation controls, whereas we want to render them before. This is why I chose to use an entirely new hook.

At first I considered reusing an existing twig template, eg [`svelte.twig`](https://github.com/datawrapper/datawrapper/blob/master/templates/svelte.twig) instead of creating a separate new one. I will see in the future if it's possible to combine them, but for now, for the sake of simplicity, I will work in a separate one, using only the variables that are needed. This way I can also work on new annotation controls separately without needing to worry about compatibility with other parts of the app that use `svelte.twig`.

---

Since none of the included changes affect existing code, this can be merged straight away.

I'd be happy to hear your ideas if you think there is a better way to deal with the hook or templates for new annotation controls.


davidkokkelink:
The alternative would be to refactor `custom_annotation_controls` that it passes in an `order` and a `render` function to display the control, so arbitrary ordering combinations can be used.

But this entire logic will be replaced when we make the switch to a svelte-based chart editor, so I don't think perfectionizing our twig codebase is worth the effort :)

ilokhov:
> But this entire logic will be replaced when we make the switch to a svelte-based chart editor, so I don't think perfectionizing our twig codebase is worth the effort :)

Yes, spending too much time on this would be somewhat counter-productive.

I'll merge this now and will continue updating `svelte-annotate.twig` as needed.

PR #348, opened 2020-05-25
Update externalData with correct id when duplicating chart

elanals:


elanals:
Thanks @davidkokkelink for the input makes much more sense. I created a PR in `external-data` https://github.com/datawrapper/plugin-external-data/pull/2 and will close this one.

PR #360, opened 2020-06-16
add access_token table and use it for authentication

gka:


migrations/013-add-access-token.sql
```diff
@@ -0,0 +1,25 @@
...
+    `type` varchar(100) NOT NULL,
+    `data` json DEFAULT NULL,
+    `created_at` datetime NOT NULL,
+    `last_used_at` datetime DEFAULT NULL,
+    PRIMARY KEY (`id`),
+    UNIQUE KEY `access_token_token_IDX` (`token`) USING BTREE,
+    KEY `access_token_user_id_IDX` (`user_id`) USING BTREE,
+    KEY `access_token_type_IDX` (`type`) USING BTREE,
+    CONSTRAINT `access_token_FK` FOREIGN KEY (`user_id`) REFERENCES `user` (`id`)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4
```
davidkokkelink, line 17:
```suggestion
) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4;
```

auto migration fails without this

lib/session/DatawrapperSession.php
```diff
@@ -56,13 +56,18 @@ protected function initUser() {
...
-                if ($auth) {
-                    $this->user = $auth->getUser();
-                    $this->method = 'token';
-                    $auth->use();
-                    return;
+            if ($authHeader[0] == 'Bearer' && !empty($authHeader[1])) {
+                $pdo = Propel::getConnection();
+                $res = $pdo->query('SELECT user_id FROM access_token WHERE `type` = "api-token" AND token = '.$pdo->quote($authHeader[1]));
+                $user_id = $res->fetchColumn(0);
+                if ($user_id !== false) {
```
davidkokkelink, line 63:
```suggestion
                $stmt = $pdo->prepare('SELECT user_id FROM access_token WHERE `type` = "api-token" AND token = :token');
                $stmt->bindParam(':token', $authHeader[1]);
                $stmt->execute();
                $res = $stmt->fetch();

                if ($res && !empty($res['user_id'])) {
                    $user_id = $res['user_id'];
```

lib/session/DatawrapperSession.php
```diff
@@ -56,13 +56,18 @@ protected function initUser() {
...
+                    $user = UserQuery::create()->findPK($user_id);
+                    if ($user) {
+                        $this->user = $user;
+                        $this->method = 'token';
+                        $pdo->query('UPDATE access_token SET last_used_at = NOW() WHERE type = "api-token" AND user_id = '.intval($user->getId()).' AND token = '.$pdo->quote($authHeader[1]));
```
davidkokkelink, line 68:
```suggestion
                        $stmt = $pdo->prepare('UPDATE access_token SET last_used_at = NOW() WHERE type = "api-token" AND user_id = :userId AND token = :token');
                        $stmt->bindParam(':userId', $user_id);
                        $stmt->bindParam(':token', $authHeader[1]);
                        $stmt->execute();
```

gka:
Yea I was thinking about prepared statements, but I thought the only benefit is performance if the query is executed multiple times per DB session (which it isn't in our case). This is from PHP's documentation about prepared statements:
>  Such a separation sometimes considered as the only security feature to prevent SQL injection, but the **same degree of security can be achieved with non-prepared statements**, if all the values are formatted correctly. 

Likewise, there was a discussion about using prepared statements in Sequelize which was ultimately rejected because despite the "feeling" they don't bring any security benefit to properly escaped values.

gka:
Having said all this, we can also use them if you want ;-)

davidkokkelink:
Yeah I'd say let's use them. The sentence after the one that you quoted from the docs:
> It should be noted that correct formatting is not the same as escaping and involves more logic than simple escaping. Thus, prepared statements are simply a more convenient and less error-prone approach to this element of database security.

To me the main benefit of prepared statements is that you don't have to think about formatting/escaping, and therefor you can't screw it up. This might be less of an issue for Sequelize core developers, who are (I hope) all very proficient with SQL, but given that we have variety of specialties among our developers it's just more difficult to know if using `intval` is sufficient, whether to use `PDO::quote`, `mysql_escape_string`, `mysql_real_escape_string`, `mysqli::real_escape_string` or something else, etc

gka:
I also added in a check for `$user->isActivated()` to make sure pending users can't use api tokens (in case they still have some)

PR #362, opened 2020-06-22
Clean up src directory, remove unused code, unify build system

davidkokkelink:
This PR:
- removes the old 'svelte 1.9' based build system in `src/rollup.config.js` (svelte 1 with svelte 2 parser mode enabled), used for the `upload` and `highlight` svelte apps, and builds them with the `src/v2` build step instead. this took surprisingly few changes to work, essentially just replacing `this.observe` with `onstate`.
- removes the remaining  removes the legacy directories `src/controls` and `src/editor`, which found a new home in `@datawrapper/controls`
- moves the `resizer` app, which had its own build system into `src/v2` as well

no user-facing changes


davidkokkelink:
now the `src` directory is empty except for the `v2` directory, so we could just move everything into `src` directly. i didn't do it yet to make this PR not completely unreadable.

gka:
would be nice to clean up the root `package.json` too, which includes loads of packages only to run the build that this PR removes.

davidkokkelink:
As requested, I have now cleaned up the `package.json` and moved the contents of the `src/v2` directory into `src` directly.

jakubdatawrapper:
I waive my review rights. But I'm very happy you made this change!

gka:
Nice. I think you're cleanup went a little too far, however. We should keep

- babel-eslint
- eslint-plugin-html
- healthier
- husky
- lint-staged
- prettier
- prettier-plugin-svelte
- watch

Not sure what's going on with the `test-legacy` script, it seems there are no more legacy tests? If I'm wrong we also need to keep `mocha` and `reify`. Otherwise the `test-legacy` and `pretest-legacy` scripts can go as well.

davidkokkelink:
I added the packages to `package.json` and also changed some components to use `httpReq.js` instead of `postJSON` to fix a bug that currently prevents team creation

gka:
linting is still not running through :/

davidkokkelink:
it works for me now, also added a `.prettierignore` instead of the old lint ignore patterns

gka:
it needs to run in circleci as otherwise we're making the circleci tests useless.

davidkokkelink:
sigh, fails with `prettier requires at least version 10.13.0 of Node, please upgrade`, and the node version i guess comes from the `sto3psl/dw-ci` container image? cc @sto3psl

gka:
@sto3psl already made a newer image, just set image to `sto3psl/dw-ci:1.4.0`

sto3psl:
yep. What @gka said.

.circleci/config.yml
```diff
@@ -1,20 +1,20 @@
...
-              - source-v1-{{ .Branch }}-
-              - source-v1-
+      - &restore_source
+        restore_cache:
+            keys:
+                - source-v1-{{ .Branch }}-{{ .Revision }}
+                - source-v1-{{ .Branch }}-
+                - source-v1-
 
-version: 2
```
gka, line 17: I think somehow the indentation got screwed up here ^

davidkokkelink: yeah looks like it

davidkokkelink:
looks like we're done here?

PR #364, opened 2020-06-30
Remove computeEmbedHeights function from publish process

elanals:
Previously `publish.embed-heights` was needed for the embed code, this is no longer the case. The height calculations performed here are incorrect, causing publish to get stuck for some locator maps due to temporary resize that happens to the preview before publish. This can be observed in [this chart](https://app.datawrapper.de/edit/yOX2F/publish), and is documented in [this ticket](https://www.notion.so/datawrapper/Certain-locator-maps-don-t-publish-code-hangs-somewhere-before-the-API-is-even-called-5cf7557fa3404f4ca5d5b50d020ddb9b).

Given that this function produces incorrect results, and the `embed-heights` property isn't used for anything, seems like we can safely remove it.


davidkokkelink:
Wonder if @gka has any other thoughts

src/account/EditProfile.html
```diff
@@ -161,9 +161,7 @@ <h2 style="margin-bottom: 20px;">{ __("account / delete / hed") }</h2>
         {:elseif deleteAccount}
         <h3><i class="fa fa-times"></i> { __("account / confirm-account-deletion") }</h3>
         <button on:click="set({deleteAccount: false})" class="btn btn-back btn-primary">
-            <!-- prettier-ignore -->
-            <i class="fa fa-chevron-left"></i> &nbsp; { __("account / confirm-account-deletion /
-            no") }
+            // prettier-ignore <i class="fa fa-chevron-left"></i> &nbsp; { __("account / confirm-account-deletion / no") }
```
davidkokkelink, line 164: this is nonsense though

elanals: I did it because I saw the same in the same file [here](https://github.com/datawrapper/datawrapper/blob/master/src/account/EditProfile.html#L102)

and the previous `<!-- prettier-ignore -->` wasn't working and that line kept getting line broken so I assumed it was something special. i wanted to run prettier to check but it changed a bunch of files so instead i just went on a limb

davidkokkelink: yeah then that is wrong too! `// prettier-ignore` will just be displayed in the UI. and i agree that the code formatter is dumb in these cases

elanals: i mean i thought it was weird but I assumed there'd be a reason. But the actual `<!-- prettier-ignore -->` wasn't working, and the line kept getting formatted, and the fact that i committed it without the line break is the reason that tests are failling

elanals: so why isn't `<!--prettier-ignore -->` working?

davidkokkelink:
verbal approval by @gka for the removal has been given in a dm to me :)

src/account/EditProfile.html
davidkokkelink: no idea, but you can just split up the line to make it shorter to make it shut up

PR #366, opened 2020-06-30
Use V3 api to save chart data

davidkokkelink:
This PR changes step 1 in Datawrapper to call the V3 API using `httpReq` to save chart data, rather than the PHP API.


gka:
did you test this in IE11? asking because of our async await friends

davidkokkelink:
Yes it works in IE11, I guess `babel-plugin-transform-async-to-promises` takes care of it

PR #374, opened 2020-07-10
Option to copy embed code directly from chart modal

elanals:
Extracted from [this dead PR](https://github.com/datawrapper/datawrapper/pull/326). Made success message more specific so user knows what they last clicked.


locale/de_DE.json
```diff
@@ -650,6 +652,7 @@
   "Layout": "Layout",
   "No charts here, yet": "Hier sind noch keine Diagramme…",
   "Why not <a href='/chart/create'>create a new chart now</a>?": "Wieso erstellst du nicht ein <a href='/chart/create'>neues Diagramm</a>?",
+  "copy-success": "der %embed-code-type% Code wurde kopiert!",
```
davidkokkelink, line 655:
```suggestion
  "copy-success": "Der %embed-code-type% Code wurde kopiert!",
```

elanals:
Yes, I'd already had the same thought and added that functionality as well, just hadn't pushed. Not sure if this approach (save to window object) was the best way to access the required team settings from the myCharts modal?

**What's new here:**
- Default/preferred embed code is indicated, and displayed at top of select
- Title of custom embed is shown instead of just 'custom'

**Possible further UI considerations, maybe @davidw-dw should weigh in:**
- How should the default be indicated, the word 'default'? 'preferred'?, what in German? Or something else, like an icon (a star or similar?)
- Does it make sense for someone have to click twice plus carefully move the cursor in order to copy the default embed which is likely what they'd want to copy in the vast majority of cases? maybe this should be radio buttons, with the default preselected or something instead?

Updated GIF:

davidkokkelink:
Great - I think as long as the default is the first choice it's fine here. With the new, no longer modal-based approach we have more space to align this with the normal publish & embed UI.

davidw-dw:
Yep, i think it's okay here. If still possible: Can we add a little bit more top-margin to the "copied.." text and reduce the line-height a little?

elanals:
yes @davidw-dw I'll do that. Do you have any opinions as to how the default option should be indicated? If we go for the text `(default)` in english, then what should it be in german? Or do you have a different idea?

davidw-dw:
I think **(default)** (and **(Standard)** probably?) are totally fine.
Maybe we can tone down it a bit by using gray text-color and reduce the size a bit. 
Screenshot 2020-07-13 at 14 36 56

elanals:
Good idea, thanks will add that!

PR #376, opened 2020-07-14
Migrate after dataset

elanals:
Possible solution for a problem in a migration for table column settings. See [this PR](https://github.com/datawrapper/plugin-tables/pull/37).


elanals:
The table PR needs to be merged so that Gregor can start working on the heatmaps, and this is still the only missing piece. Are there any reasons not to change where the migration happens here too?

gka:
the only thing that comes to my mind is that we need to test this thoroughly. So far the dataset-migration has only been used in `d3-maps`, and since this PR would bring this change to all visualizations with svelte controls we should make sure that they all work as expected.

davidkokkelink:
I can test some scenarios and if everything's fine we can merge later today

gka:
thanks

davidkokkelink (approved):
looks good :+1:

davidkokkelink:
resolved with svelte visualize

PR #380, opened 2020-07-15
Less intimidating display of null values in check & describe

elanals:
Consider this a feature request in the form of a PR.

Highlighting empty cells as red feels misleading because it communicates 'there's a problem here' which I don't think is the case, just because there's no value in a cell. I noticed the reaction of fear and concern in trainings I gave.

Printing the word 'null' also doesn't really seem to add much value, but rather look a bit confusing and unfamiliar.

Thoughts?

Before:
image

After:
image


gka:
fine with me!

davidw-dw:
Hey, thanks for PR. Good idea!
To be honest i'm not sure about dropping both indicators (color + null). I'd agree it can be intimidating the way it is now and _null_ is probably not very clear, but I always liked the quick overview that this distinction creates. I'd assume that in most of the cases you don't want / have single empty cells within a one column. When looking at the _after_ version i tend to control  each line in column D + E, whereas with the _before_ version i have the feeling that there are no gaps - otherwise i would have seen them already after learning this pattern.

What do you think about that? Is it just me?
I was thinking about dropping just one of the indicators (probably null) or use a new one that doesnt look like error but still creates the distinction between "complete" and "missing".
Example with just red bg color:
Screenshot 2020-07-16 at 09 07 29

elanals:
Thanks for taking a look @davidw-dw. I created a [ticket](https://www.notion.so/datawrapper/Reconsider-display-of-empty-cells-in-Check-Describe-0d4ebde524ca41c08dd25ee7959e9e46) to continue the conversation.

gka:
I converted this to draft to snooze notifications until the discussion is resolved

elanals:
The new look:
image

PR #386, opened 2020-08-10
add simple dw.backend.hooks interface

gka:
this is a very quick and simple implementation of a client-side hook system
```js
// register hooks
dw.backend.hooks.register('meaning-of', (what) => {
    return what === 'life' ? 42 : 'nobody knows';
});
// one more hook registrar
dw.backend.hooks.register('meaning-of', (what) => {
    if (what === 'life') return 'to be happy';
    throw new Error('not implemented');
});

// call hooks
dw.backend.hooks.call('meaning-of', ['life']);
// errors: Array []
​// results: Array [ 42, "to be happy" ]

dw.backend.hooks.call('meaning-of', ['everything']);
// errors: (1) […]
​​//   0: Error: not implemented
​​// results: Array [ "nobody knows" ]
```

there are also a few minor changes in this PR:
- add new `VISUALIZE_BEFORE_SVELTE_SIDEBAR` hook to insert twig into the svelte-sidebar.twig
- pass `userThemes` to visualize template so it can be used in `svelte-sidebar.twig`


gka:
you can see the hooks in use [here]](https://github.com/datawrapper/plugin-chart-locale-select/pull/2/files#diff-dcd6aa7812e58510f5be3a4273b57213) and [here](https://github.com/datawrapper/plugin-d3-maps/blob/color-scale/src/controls/shared/Design.html#L20-L22)

PR #389, opened 2020-08-20
remove print-mode from resizer app

gka:
I believe this isn't in use anymore, and also it's broken because our new API doesn't implement the `/v3/charts/{id}/print` endpoint. so let's remove this code.


gka:
closing, since this apparently is still in use

PR #390, opened 2020-08-20
Replace old colorblind check with Svelte component

gka:
Thought it was about time to refactor the colorblind check.


davidkokkelink:
wonderful! will give these a deuteranomalous review shortly

PR #393, opened 2020-08-25
Fix undef vars in www/static/js/dw and check it with npm run lint

jakubdatawrapper:
#### Motivation

We have a pre-commit hook that does static code analysis (lint) of `www/static/js/dw` using healthier (eslint) and prettier. However this code doesn't pass the check. Which means that if I make a change in `www/static/js/dw` I can't commit the code unless I fix all the errors (which I didn't cause).

#### Changes

- Fix functional errors in the code:
    - undefined variables
    - multiline strings (supported only in ES > 2015)
    - undefined global variables
- Ignore style errors (configure eslint to only print warnings).
- Add the code to the `npm run lint` check.

#### Remaining issues

Now the code passes the `healthier` part of the pre-commit hook but still doesn't pass the `prettier` part. What needs to be done is to reformat the code with `prettier`. Since this will change most of the lines of code and would make this PR hard to read, I will do that in a separate PR.


gka (approved):
looks good to me

PR #395, opened 2020-08-26
Call API with CSRF protection headers

jakubdatawrapper:
#### Motivation

Required by https://github.com/datawrapper/api/pull/215

#### Changes

- `call_v3_api`: Set CSRF token header and Referer

- `dw/backend`: Set CSRF token header when calling the API

    Add a module `httpReq`, which is a jQuery-based implementation of datawrapper/shared/httpReq, compatible with IE 11.

- Make the code compatible with current version of datawrapper/shared:

    Because of modules that were moved to chart-core:

    - chart

#### Testing

- [X] Test `call_v3_api`. -> Done, because the page crashes immediatelly when it doesn't work.
- [X] Test chart data syncing. -> Done, because it happens immediatelly when the chart editor is loaded.
- [x] Test chart duplication link in "this chart belongs to" alert.
- [x] Test chart duplication link on the Publish step.
- [x] Test chart duplication link in the chart dropdown on Team Chats page.
- [x] Test chart duplication link in the chart modal window.
- [x] Test chart duplication link on User Detail page.
- [x] Test logout link.
- [x] Test changing of embed type on the Publish step (tests `setUserData()`).
- [ ] Test that the chart editor loads in IE 11.


jakubdatawrapper:
Please review @davidkokkelink 

The test scenarios I tried are mentioned in the PR description.

If you could test it also in IE11, that would be great!

davidkokkelink:
Went through your checklist on IE11 and it works fine, and the `crumb` cookie is set in requests:

crumb

I wasn't sure about these two items:
- _Test chart duplication link on User Detail page_
-> what is this referring to?
- _Test changing of embed type on the Publish step (tests setUserData())._
-> the functionality worked but switchting the embed type on `/publish` didn't seem to trigger an HTTP request for me?

davidkokkelink:
Yes, I think that lives in [`plugin-admin-chart`](https://github.com/datawrapper/plugin-admin-chart). But if Datawrapper admins end up using IE11 we probably have bigger problems at our hands

davidkokkelink (approved):

jakubdatawrapper:
> the functionality worked but switchting the embed type on /publish didn't seem to trigger an HTTP request for me?

I checked it and the app was actually sending a successful request only the first time embed type changed. On subsequent changes, the requests were failing with error 404, because we were sending an invalid request payload. I've fixed it.

PR #398, opened 2020-09-02
Fix crash when browsing a folder in my charts

jakubdatawrapper:
#### Steps to reproduce

1. Go to https://app.staging.datawrapper.de/mycharts
2. Click on a folder in the folder tree.
3. See that the page is blurred and there is an error in the console that variable `cft` is undefined.

Doesn't happen on production.

2020-09-02 12-55-21 1501x849

#### Changes

I guess `cft` should be a reference to `window.ChartFolderTree`. Using `window.ChartFolderTree` seems to fix the problem. However if I do `window.ChartFolderTree === cft` on production it's false, so the two objects are not identical.

I might have broken this in #393 or #394.


EX0l0N (commented):
Wew. This is a tough one. I remember introducing the `cft` object for a specific reason, but I don't remember what it was.
So while your patch might fix a crash because the `cft` does not exist, I won't expect those to be interchangeable.

I need to dig out why the `cft` object was created in the first place, before I can give you green light for this.

Atm my construction site is so far away from this, that it's hidden behind the horizon. I need to bring my own project to a point, where dropping my _in head_ stack and registers won't waste hours of work.

I'm trying hard to review this this week, but I can't promise anything.

davidkokkelink:
it seems obvious that this is caused by [this erroneous clean-up](https://github.com/datawrapper/datawrapper/pull/393/files#diff-da2ff5558c4d7fb1ff6434bf317793b6L24), as linked by jakub. let's not spend too much time re-immersing into this code base

davidkokkelink (approved):

jakubdatawrapper:
I think I found the cause of the problem. I removed an assignment to `cft` [here](https://github.com/datawrapper/datawrapper/pull/393/files#diff-da2ff5558c4d7fb1ff6434bf317793b6) because I thought it's just an unused variable, while in fact it was assignment to `window.cft`.

EX0l0N (commented):
All good then. Tnx for clearing this up so fast.
👍

PR #399, opened 2020-09-02
Fix XSS on Settings and in My Charts drag & drop

jakubdatawrapper:
#### How to reproduce

1. Settings:

    2020-09-02 14-50-58 1200x559

2. My Charts drag & drop:

    2020-09-02 14-06-34 2560x1440

Issue: https://www.notion.so/datawrapper/XSS-in-notification-all-team-members-can-access-chart-978ea1e7b10248bb992c714f86c11de0

#### Changes

Escape team name.

Add function `escapeHTML()` for that reason, which is -- quite honestly -- a snippet from Stack Overflow. If this implementation is fine, I will move this function to datawrapper/shared in a follow-up PR.


davidkokkelink (approved):
interesting approach! i think we still want to use the existing `purifyHtml` function in `@datawrapper/shared` due to its customizability & whitelisting capabilities, but this should do the trick for the not-yet-rolluped parts of our code.

jakubdatawrapper:
The difference is that for a team named `<b>my team</b>`, `purifyHtml` would print just `my team` while `escapeHTML` will print `&lt;b&gt;my team...`.

PR #401, opened 2020-09-03
Fix svelte sidebar updates

davidkokkelink:
Some minor fixes to get in-chart live updates and some other functionality working properly again


gka (approved):
Thanks

PR #403, opened 2020-09-07
Use _.isEqual

davidkokkelink:
Using `_.isEqual` instead of comparing `JSON.stringify` fixes a bug that caused visualizations to be re-rendered needlessly when the _order of keys_ in an object changed.

Underscore is currently still a globally [available dependency from core.twig](https://github.com/datawrapper/datawrapper/blob/is-equal/templates/core.twig#L54), so should be fine to use here.


gka (approved):
nice, didn't know about `_.isEqual`!

PR #404, opened 2020-09-14
Add redis client integration

davidkokkelink:
This PR allows us to make calls to `redis` from PHP. The config lives in `config.yaml`, at the global scope:

```
redis:
  host:
  port: 
  password: 
```

To use it somewhere:

```php
if (Redis::isInitialized()) {
  $val = Redis::get('some-key');
}
```

Might require running `composer dumpautoload` for the new file to be picked up by `composer`.


lib/bootstrap.php
```diff
@@ -27,6 +27,10 @@
     }
 }
 
+if (isset($dw_config['redis']) && !empty($dw_config['redis']['host'])) {
+    init_redis($dw_config['redis']['host'], $dw_config['redis']['port'], $dw_config['redis']['password']);
```
gka, line 31: since the Redis class is global as well I think we can ditch the one-liner init function

```suggestion
    Redis::init($host, $port, $password);
```

lib/utils/init_redis.php
```diff
@@ -0,0 +1,41 @@
+<?php
+
+function init_redis($host, $port, $password) {
+	Redis::init($host, $port, $password);
+}
+
```
gka, line 6:
```suggestion
```

gka (approved):
looks good, though I would ditch the `init_redis` function

davidkokkelink:
yeah i only threw it in to keep the consistency with the filename (filename == function name which is apparently convention in `lib/utils`). but i don't really care!

PR #408, opened 2020-09-24
Hide scrollbar from the chart preview iframe

jakubdatawrapper:
To make the preview match the print export, which doesn't render scrollbars.

Relates to: https://github.com/datawrapper/plugin-export-pdf/pull/157


jakubdatawrapper:
Is there a use case where we would want scrollbars in the preview, @davidkokkelink?

davidkokkelink (approved):

PR #409, opened 2020-09-25
add client-side cache for localization messages 

gka:
Since we moved to client-side templating using Svelte we've been sending our localization messages to the client with each request. This PR aims to optimize this using `localStorage`.

Works as follows:
* server computes a md5 hash of the entire client-side messages object and sends it to the client
* client checks if the hash is "new" 
* if it's new it stores this hash along with the entire messages object in `localStorage`. we also set a cookie `DW-MESSAGES-HASH` to let the server know we have cached this version for future requests
* if it's not new, and we already have the hash, we're loading the messages object from `localStorage`

for the next request
* the server can now check the client-side hash from the cookie against the freshly computed hash
* if they are identical, we send an empty object instead of the messages object

This saves us ~154k of data per request.


davidkokkelink:
nice! saves about 40KB gzipped which for most pages is >50% (e.g. dashboard went from 55kb gzipped to 11, chart upload page from 65 to 21, etc.)

davidkokkelink (approved):

PR #410, opened 2020-09-25
Include dataset in migration

davidkokkelink:
Pass the dataset to the migration code and reapply changes after the migration has been performed.


gka:
there is a good reason for applying the migration _before_ applying the visualization-defined default metadata.visualize settings. Migration code often only runs if a certain key is missing. If the defaults are applied before the migration, those keys will already be set and the migration won't run.

davidkokkelink:
valid concern, and i'm happy to look at a better solution in the context of the svelte visualize app. but as `svelte-sidebar` will only be used temporarily for d3-maps, and their migration works only by checking for the existence of old metadata keys, i think we are good to launch like this.

gka (approved):
alright I think it should work now. can you please test your migration code again before merging?

davidkokkelink:
yes, works now :+1: if you approve `chart-core` PR i will merge this

PR #411, opened 2020-10-02
Perform polyfill check on the client

davidkokkelink:
By doing [this check](https://github.com/datawrapper/datawrapper/blob/master/controller/polyfill.php#L7-L12) on the client first, we can shave about 500ms (!) from every page load in Datawrapper.

Tested in IE11


gka:
yes that makes sense. although we could also just move this to our new frontend server while we're at it.

gka (approved):

gka:
Like, we could just do:
```js
if (....) {
    document.write('<script src="/lib/polyfills/ie-11.js"><\/script>');
}
```
and then delete the `/polyfill` PHP route.

davidkokkelink:
yeah let's do that!

PR #415, opened 2020-11-05
Clean axes object before passing it to controls

davidkokkelink:
So they don't fail when column names change


davidkokkelink:
hey @elanals what do you think about this totally original idea of mine

elanals (approved):
Create use of hasColumn, 5 stars would recommend. linting could be better though.

PR #422, opened 2020-11-12
Fix upload textarea placeholder not visible

jakubdatawrapper:
#### Steps to reproduce

1. Create new chart.
2. See that the data upload textarea is plain white instead of showing the message "Paste your copied data here..." as expected.

2020-10-21 14-03-00 1208x660

#### Problem

PHP was setting a single newline character as chart data, which made the textarea placeholder not visible, because the textarea had some content, albeit just a newline character.

#### Changes

Trim the chart data string.

Do the same in `PublicChart`, because I assume these two classes should be kept in sync.


davidkokkelink:
thx, btw the root cause of this is this behavior: https://github.com/datawrapper/plugin-chart-data-s3/blob/master/api.js#L27-L28

we added it in a hurry because it suddenly became an issue during the launch of the node-based publish-preview. it would be nice to at some point fix it so that a GET asset call on a non-existing chart just returns a 404 and all parts of our app handle it appropriately.

davidkokkelink (approved):

PR #428, opened 2020-11-26
Save team settings.controls to chart store

elanals:
So that it can be accessed directly in controls. For motivation and more details, see [this ticket](https://www.notion.so/datawrapper/Team-Feature-Customize-selection-in-Custom-format-dropdown-e36719b0bb17410481a06ecd43b94a61).


controller/chart/visualize.php
```diff
@@ -64,6 +64,8 @@
 
         $res = Hooks::execute('enable_web_to_print');
         $webToPrint = !empty($res) && $res[0] === true;
+        $org = $chart->getOrganization();
+        if ($org) $teamSettingsControls = $org->getSettings("controls") ?? [];
```
gka, line 68:
```suggestion
        if ($org) $teamSettingsControls = $org->getSettings("controls") ?? new stdClass();
```

gka (commented):
I think I would just name it `teamSettings`. As far as the chart editor is concerned, that's the only available team setting. If we want to keep it open for more team settings in the future, we can also pass the "full" object structure, but leave only the `controls` key. This way the team settings "paths" would be the same.

```js
teamSettings: { controls: teamSettingsControls }
```

but apart from that it's fine.

davidkokkelink (approved):

elanals:
@gka, I made changes, can you confirm that this is what you going for?

gka (approved):
looks good

PR #430, opened 2020-12-01
Two-factor authentication + user settings

gka:
see also https://github.com/datawrapper/api/pull/261

This PR adds a rudimentary two-factor authentication using YubiKey devices. Rudimentary because I didn't use the fancy browser `u2f` API (which would show the nice "insert your U2F device" prompt near the address bar) but a dumb password field with autofocus.

Let's start with the user settings where (currently only admins) can activate a Yubikey. The page requests a list of available OTP providers via `GET /v3/me/otp` (see corresponding [API PR](https://github.com/datawrapper/api/pull/261)). The route also returns if a provider has been enabled. Then the user gets a button to activate a provider. After clicking, the user needs to provide a valid OTP token so we can associate the device with the users account (more about the details in the API PR). Once enabled, the user can decide to disable the OTP provider (in case a device gets lost). In the current implementation we're not asking for a final valid token from the old device, but we could change that if we wanted.

otp-settings

Here's how the login looks once the OTP has been enabled:

otp-login

Under the hood the form sends a normal login request, and if the response is a `401` with the message `Need OTP` the form will show the OTP prompt. Should we add more OTP providers (Google Authenticator etc), we should letusers with multiple enabled OTP providers decide which one they want to use.

My idea was to launch this as an admin-only feature for now, because it adds security as it is. 

Later we can polish the UI (translation, UX feedback etc) and also switch to the `window.u2f` browser API which I think is more in line with what users expect.


davidkokkelink (approved):
Cool! I think it's a good addition that adds security for admins asap.

If we make it available to users at some point we need to think about things like recovery tokens, but for now I'd say let's merge it and enable it for our admin accounts.

PR #431, opened 2020-12-02
add support for authenticator otp

gka:
depends on https://github.com/datawrapper/api/pull/262

again, the UI is pretty rough:

otp-authenticator

If a user has set up multiple OTP providers, the signin will pass if at least one of them succeeds with the provided OTP token.


davidkokkelink (approved):
Interesting, nice to see that the QR code generation is so straightforward

PR #436, opened 2020-12-11
add support for options.data in jquery-based httpReq clone

gka:
otherwise it's impossible to send a PUT request with a plain text body, e.g., to set csv data to /v3/charts/:id/data


jakubdatawrapper (approved):
Looks good to me

PR #438, opened 2021-01-07
Initialize chart locale in visualize step

elanals:
Pie chart controls call `vis.chart().local()` to determine which default translated text to prefill the pie slice Group, and donut Total label inputs with ([here](https://github.com/datawrapper/plugin-d3-pies/blob/06fe5d146f12aa52171919853f791ccf889a0e76/src/controls/Controls.html#L466-L470)).

Currently this is undefined, as `chart().local()` only returns a value if it has previously been initialized, which currently doesn't happen on step 3, only in all other steps:
https://github.com/datawrapper/datawrapper/blob/140bf4809eb94932018c0deead7efada65c1c470/templates/chart/base.twig#L44-L54

This PR adds the locale initialization to visualize's main.js. The fallback is set to `en-US`, mirroring the behaviour of the preview [here](https://github.com/datawrapper/frontend/blob/cf02fd827236b1b7808e7e1041e327ab469eb991/src/routes/preview/index.js#L75).


gka (approved):
looks good to me :+1:

PR #441, opened 2021-01-15
Refactor controls

sjockers:
## Changes

- Install @datawrapper/controls v2.34.0
- Update Svelte modules to work with @datawrapper/controls v2.34.0


gka (approved):
so good to get rid of these `<ProductTable />` and `<MemberTable />` confusion now.

PR #442, opened 2021-01-22
Improve publish page css at lower breakpoint

elanals:
This PR adds some css to handle the appearance of the publish step under the 979px breakpoint, which currently looks like this:

Screenshot 2021-01-22 at 18 18 14

This is relevant primarily for customers using Datawrapper via CMS integration, who typically display the app in a smaller window contained within their CMS.

Specifically I made the publish button, and the action buttons smaller.

I also took the liberty of adding some margin above the publish alert.

image


gka (approved):
so glad you took care of this! :+1:

PR #444, opened 2021-02-02
Get presets from the correct place in the theme

elanals:
### Motivation
This moved from `theme.export.presets.pdf` to `theme.export.pdf.prests`, but that change wasn't implemented in `Resizer.html`, which means that print presets wouldn't be shown.

### Changes
Get from correct place in theme - from `$themeData`, rather than `theme` (as far as I can tell `theme` was just permanently null, not sure how/if this ever worked, as it's not bound to the component?)
https://github.com/datawrapper/datawrapper/blob/8aaf80bab5ef4b8b949c3185ebe9361679b56b01/src/visualize/App.html#L18

Also condensed the code a bit.


gka (approved):
let's hope `export.pdf.presets` never ends up getting set as string or number, as in this case the `presets.length` will cause a JS exception. personally I'd rather be safe than sorry and keep something like 
```js
hasPresets({ presets }) {
    return presets && presets.length;
}
```

elanals:
This code already ensures that presets is an empty array if it's unset/null/undefined. And I think it's safe enough to assume that it won't ever be a number or anything invalid because we have an enforced theme schema, so I will merge this.

PR #445, opened 2021-02-07
Fix developer URL in README

bbenno:


davidkokkelink:
Thanks for the report, this has been fixed in the meantime.

PR #446, opened 2021-02-11
Save chart before writing data

davidkokkelink:
Otherwise the Node API won't know that the user has access


gka (approved):

EX0l0N:
I don't understand what that specific file is doing in the big picture (because I didn't take the time to read into all of that), but I'd assume moving that instruction further up won't make anything worse than it was before.

It looks like a legit solution of a _this needs to happen before that_ issue.

EX0l0N (approved):

PR #447, opened 2021-02-13
Get csrf crumb if needed

davidkokkelink:
There are usage scenarios in which PHP's `call_v3_api` is the first time that a session-authenticated POST request is made to the V3 API. In this case, we need to make a GET request to the API to get a CSRF crumb similar to how it is performed in the client-side `httpReq` function.


jakubdatawrapper (approved):
Makes sense and code looks good, thanks.

PR #454, opened 2021-03-22
Use external metadata file in step 3

davidkokkelink:
When opening a visualization with external metadata in step 3, load external metadata so that the UI can use it to determine which inputs to disable.

Related: https://github.com/datawrapper/controls/pull/158


controller/chart/visualize.php
```diff
@@ -90,6 +90,17 @@
                     return ['id'=>$t->getId(), 'title'=>$t->getTitle()];
                 }, ThemeQuery::create()->allThemesForUser($chart)),
         );
+
+        try {
+            if ($chart->getMetadata('data.external-metadata')) {
```
gka, line 95: in the other PR you also checked that upload-method is "external-data". Guess it would be good to be consistent with the conditions.

davidkokkelink: makes sense, updated

PR #455, opened 2021-03-31
Set session cookie in $_COOKIE if it doesn't exist already

davidkokkelink:
When a guest user accesses Datawrapper, we generate a session. On subsequent requests, this session is available in the superglobal `$_COOKIE`, which contains the cookies the _user request_ sent. However, on the request that causes the session to be created in the first place, `$_COOKIE` remains empty even after the session has been created.

This breaks functionality that expects that session cookie, most importantly us [calling our own V3 API](https://github.com/datawrapper/datawrapper/blob/master/lib/utils/call_v3_api.php#L22). This PR fixes it by setting the session value for the specific case that the session was just created but isn't present in `$_COOKIE` yet.


lib/session/DatawrapperSession.php
```diff
@@ -62,6 +62,8 @@ public static function initSession() {
             $cookieOpts['samesite'] = (isset($_SESSION['type']) && $_SESSION['type']
                 == 'token') ? 'None' : $sameSite;
             setcookie($ses, $_COOKIE[$ses], $cookieOpts);
+        } else {
+            $_COOKIE[$ses] = session_id();
```
EX0l0N, line 66: I've read up a little bit by now and assuming that `session_id()` will always hold a valid ID (and i guess that's what line 56 ensures), this should be working.

Hmm.

If a session is present when we hit line 59 we do things to the cookie (expires, samesite…).
Now, if the session was not present, we set it - but we don't do the above transformations to the cookie in that case.

Is this what we want?

davidkokkelink: Yes. Keep in mind that transformation doesn't alter the `$_COOKIE` object in any way, it only uses that to determine the presence of a request cookie, and in turn alters the _response cookie_  via `setcookie`. We don't need to alter the response cookie in any way here.

PR #456, opened 2021-04-13
Don't crash on empty or conflicting computed column name

jakubdatawrapper:
#### Steps to reproduce

1. Add a new computed column.
2. Clear the _Column name_ field.
3. Type in a _Formula_ and wait a second.
4. See that an exception appears in the console.

    Moreover, after refreshing the page there is another exception in the console which makes it impossible to do anything with the table.

    There is also a similar error when renaming a computed column so it has the same name as a normal column.

2021-04-13 14-31-26 1241x852

Issue: https://www.notion.so/datawrapper/Impossible-to-remove-computed-column-after-adding-normal-column-of-the-same-name-c7e1455fcd5541088b16b08d669e5f5e

#### Problem

The problem is that a column with an empty or conflicting name gets renamed to `X.1` or `Liquid fuel.1` or so by chart-core. Then when we try to find a computed column called `X.1` or `Liquid fuel.1` in `metadata.describe.computed-columns`, it doesn't exist (because in the metadata, it's called `""` or `"Liquid fuel"`) and everything breaks.

#### Solution

Use `Column.origName()` when finding computed column in chart metedata.

#### Renaming issues

- [x] Depends on https://github.com/datawrapper/chart-core/pull/104
- [x] Creating two computed columns with the same name merges them into one (because there is no conflict resolution logic when saving `metadata.describe.computed-columns`). -> Is it okay? Can we leave this for now?
- [x] I removed the option to add a variable for normal column into a formula of a computed column with the same name, because chart-core treats that as a circular dependency. -> Is that okay too?
- [x] I removed a try..catch block from the computed property `columns` in App.html, because I don't see when that would throw an exception.


ilokhov:
The code looks fine to me but I don't know enough about computed columns to answer your questions, so I would suggest requesting a review from someone like @gka.

Btw I couldn't trigger an error by clearing the column name but only when using a duplicate name (used the same demo dataset as you)

gka (approved):
>  Creating two computed columns with the same name merges them into one (because there is no conflict resolution logic when saving metadata.describe.computed-columns). -> Is it okay? Can we leave this for now?

yes, that's ok for now

> I removed the option to add a variable for normal column into a formula of a computed column with the same name, because chart-core treats that as a circular dependency. -> Is that okay too?

so if I create a computed column named `foo` I can't use a data column that is also named `foo`? I think that's acceptable

jakubdatawrapper:
> so if I create a computed column named foo I can't use a data column that is also named foo? I think that's acceptable
Yes, that's the scenario.

I pushed two more changes to the PR:
- Upgrade of chart-core to get the `origColumn` feature.
- A fix to avoid two computed columns with the same name in chart metadata.

I'm merging the PR now.

PR #461, opened 2021-05-19
Fix robot_500 font URL

jakubdatawrapper:
Change the URL form absolute to relative, because all other fonts have relative URL.

Not tested.

See also https://github.com/datawrapper/frontend/pull/64


jakubdatawrapper:
Merging, because it's the same as the approved https://github.com/datawrapper/frontend/pull/64

PR #465, opened 2021-05-20
Add legacy dependencies

davidkokkelink:


EX0l0N (approved):
Looks like this should do the trick.

PR #469, opened 2021-05-28
without this options curl waits (forever) for body when method is HEAD

EX0l0N:
The docs of `CURLOPT_CUSTOMREQUEST` explicitly say methods custom _other_ than `GET`& `HEAD` but it seems to work anyways.
A few lines below that they state that `GET` is an allowed value, so this is kinda misleading.

Nevertheless curl _needs_ the `CURLOPT_NOBODY` option set for `HEAD` requests because it parses the body size which is transferred on `HEAD` too and it will wait for bytes to come without it.


davidkokkelink (approved):
makes sense

gka (approved):

PR #470, opened 2021-05-28
api reachability check from core

EX0l0N:
This implements a PHP → API check, which can be used in StatusCake to alert admins if something is wrong. 

It needs https://github.com/datawrapper/datawrapper/pull/469 to work.

### DIFF
--- a/www/api/index.php
+++ b/www/api/index.php
@@ -145,7 +145,7 @@ $app->get('/status', function() use ($app) {
     ok();
 });
 
-$app->get('/v3_reachability', function() use ($app) {
+$app->get('/status-v3', function() use ($app) {
     $res = call_v3_api('HEAD', '/');
 
     $app->response()->status($res[0]);
--- a/www/api/index.php
+++ b/www/api/index.php
@@ -153,7 +153,7 @@ $app->get('/v3_reachability', function() use ($app) {
     if ($res[0] === 200) {
         ok($res[1]);
     } else {
-        error('API unreachable!', 'This needs to be fixed immediately.');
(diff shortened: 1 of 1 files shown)


www/api/index.php
```diff
@@ -145,6 +145,18 @@ function if_is_admin($callback, $elseCallback=null) {
...
 
+$app->get('/v3_reachability', function() use ($app) {
+    $res = call_v3_api('HEAD', '/');
+
+    $app->response()->status($res[0]);
+
+    if ($res[0] === 200) {
+        ok($res[1]);
+    } else {
+        error('API unreachable!', 'This needs to be fixed immediately.');
```
davidkokkelink, line 156:
```suggestion
        error('api-unreachable', 'The V3 API is unreachable.');
```

I think the convention for these `error` calls is that the first parameter should be designed to be machine readable

www/api/index.php
```diff
@@ -145,6 +145,18 @@ function if_is_admin($callback, $elseCallback=null) {
     ok();
 });
 
+$app->get('/v3_reachability', function() use ($app) {
```
davidkokkelink, line 148:
```suggestion
$app->get('/status-v3', function() use ($app) {
```
how about this name (i think we use dashes instead of underscores in url routes typically)

davidkokkelink (approved):
thanks! two wording suggestions, other PR is approved already :+1:

EX0l0N:
> thanks! two wording suggestions, other PR is approved already 👍

I know. I prefer to wait for @gka s review though. This might easily break things if not done right.

gka (approved):

PR #476, opened 2021-07-21
Show an error message when the saving of team settings fails

jakubdatawrapper:
#### How to reproduce

1. Make sure the `admins` team has no owner:

    ``` sql
    UPDATE user_organization SET organization_role = 2 WHERE user_id = 1 AND organization_id = 'admins';
    ```

    This is actually the [default](https://github.com/datawrapper/plugin-default-data/blob/master/migrations/005-add-non-privileged-user.sql) and will be changed in another PR.

2. Go to http://app.datawrapper.local/team/admins/custom-fields

3. Add a custom field.

4. See that there is an error 500 in the console but no message is shown to the user.

2021-07-21_11-03-15_1523x705

#### Changes

Catch error from `storeTeamSettings()` and show an alert box.

2021-07-21_10-51-01_854x419


EX0l0N (approved):
Nice 👌

PR #477, opened 2021-07-22
Support Svelte 3 action components

jakubdatawrapper:
#### Motivation

Svelte 3 export-pdf.

#### Changes

1. Pass the `store` as a property.

    Because Svelte 3 doesn't support global store.

2. Pass dynamic properties as a `data` property.

    This is more convenient to work with than when the properties are stored in `this.refs.action`. Keep `this.refs.action` though, so that this PR can be merged before the Svelte 3 `export-pdf` and nothing breaks.


ilokhov (approved):
Looks good, thanks!

PR #478, opened 2021-07-27
Immediately select publish action

jakubdatawrapper:
#### How to reproduce

1. Go to https://app.datawrapper.de/edit/DmCOi/publish#export-pdf
2. See that the PDF action is not selected and the PDF export form is not visible.
3. Expected: the action is selected and form visible, because there is `#export-pdf` in the URL.

#### Changes

Check `window.location.hash` when the Publish.html component is created and select the action.


ilokhov (approved):
Looks good, thanks!

PR #480, opened 2021-08-06
Chore: Upgrade dependencies and fix build warnings

sjockers:
This bumps `@datawrapper/controls` and other internal packages to the latest version. It also fixes & silences a range of build warnings that made the output of the build process largely unusable. 

The remaining warnings are actual accessibility problems that need to be fixed separately.

## Changes

- Upgrade `@datawrapper/controls` and related dependencies
- Upgrade `rollup` and related dependencies
- Setup `rollup` so that it no longer complains about packages loaded at runtime 
- Silence circular dependency warning for `d3-...` packages


gka (approved):
works fine for me, recompiled the svelte apps and tested the chart editor a bit and seems to work as before. thanks for the dependency update!

PR #481, opened 2021-08-10
Fix/faster mycharts search

gka:
* improve MyCharts search query structure and only search on chart title (better than timeouts!)
* add FULLTEXT index for `chart.title` to enable [full text boolean search in MySQL](https://dev.mysql.com/doc/refman/5.7/en/fulltext-boolean.html)


gka:
converted to draft as creating the FULLTEXT index is a non-trivial task with our database. Takes more than two hours to run while completely blocking writes to the `chart` table.

PR #482, opened 2021-08-18
Unpublish: Modal confirmation

sjockers:
Add a modal confirmation step for un-publishing charts.

TODO:
- [ ] Translate UI strings


sjockers:
Here's a demo: https://app.neurath.datawrapper.dev/chart/kPPhn/publish (unpublishing isn't implemented yet, only the modal prompt)

davidkokkelink (commented):
looks good to me! i think we can already implement the call to `POST /v3/charts/:id/unpublish` now. i'll merge it later today, and until then this gives us a good opportunity to test the error handling code path ;)

sjockers:
Thanks @davidkokkelink! Unpublishing now works, check it out: https://app.neurath.datawrapper.dev/chart/kPPhn/publish

davidkokkelink:
cool! two notes on this thing:
replace-code

- it definitely shouldn't display on the _first_ publish (which happens now)
- it's also factually untrue, and i think needlessly 'error-like'. the reality is: you _don't_ need to replace the embed code in your CMS, we redirect automatically. however, the existing version might have been cached by some of your website visitors, so you _can_ replace it if you want to guarantee the new version being visible for everyone immediately.

davidkokkelink:
i guess this is the wrong PR for this feedback though?

elanals:
Maybe something like:

> Your embedded visualization will now automatically update in a few moments. In case you're noticing a longer delay, you can replace the embed code with the latest one.

davidkokkelink:
regarding this PR, I think it's good to go once the translations are implemented. also wouldn't it make sense to implement this for both `App.html` and `Publish.html` at the same time?

elanals:
if we even want to sound negative. but then again this is a big point of confusion, letting people know they don't need to re-embed would be helpful i think

davidkokkelink:
I like @elanals' wording suggestion

sjockers:
Thanks for your feedback @elanals / @davidkokkelink: 

Yes, the alerts shown after publishing are still work in progress. The best place for discussing them would be this notion ticket: https://www.notion.so/datawrapper/Inform-user-that-they-may-need-to-update-the-embed-code-after-re-publication-4e9b542735b44ae8b5e292d993a6c33f

> regarding this PR, I think it's good to go once the translations are implemented. also wouldn't it make sense to implement this for both App.html and Publish.html at the same time?

Yes, good point. I'll do that 👍

davidkokkelink (approved):
great, thanks. approving this in case you want to handle translations and alert in the main dev branch.

davidw-dw:
The discussed message is for the case that re-directing is turned off. There is another message with a different wording and different design for normal cases. As @sjockers already pointed out this [notion ticket](https://www.notion.so/datawrapper/Inform-user-that-they-may-need-to-update-the-embed-code-after-re-publication-4e9b542735b44ae8b5e292d993a6c33f#ee2f4cc6409b4c3085be47883bdffe52) is probably better to discuss the design and wording. There both cases can be seen as well.

davidkokkelink:
ah, perfect. glad this was already handled properly, sorry for my misunderstanding.

davidw-dw:
@sjockers here is some feedback for the modal. as discussed this implementation is technically not a new standard. we'd do that in svelte3 instead.

- all text should have at least `font-weight: 400;`
- i think here is a misunderstanding regarding the button style. the button style used in the design is the one thats already been implemented with `.btn` and `btn-danger`. No need to redo this. Also hover styles get applied that way. To be honest i don't know from _where_ this styles comes. Maybe bootstrap? 
- the back btn should use `font-weight: 700;`
- we can use the new **×** icon in the icon repo for closing, also it should be a bit larger, but thats not really important. i adjusted the design in [figma](https://www.figma.com/file/pV0xUMtTfXfpc1TjQom0r4/Embed-UI?node-id=581%3A493) as well.

sjockers:
Thanks for you feedback @davidw-dw! 

> - all text should have at least font-weight: 400;

I've fixed it for the text in the modal. It looks like `font-weight: 300` is the site's default style for text paragraphs. If we never want that, it makes sense to change it for the whole site.
 
> - i think here is a misunderstanding regarding the button style. the button style used in the design is the one thats already been implemented with .btn and btn-danger. The second is used in the App > Settings for example. No need to redo this. Also hover styles get applied that way. To be honest i don't know from where this styles comes. Maybe bootstrap?
> - the back btn should use font-weight: 700;

Yes, it's slightly customized bootstrap styling. I was confused because the padding and the font props were slightly diffferent in the mockup. I'm reusing the default `.btn` styles now with adjusted padding and font weight, where necessary.

> -we can use the new × icon in the icon repo, also it should be a bit larger and brighter, but all that's not really important. i adjusted the design in figma as well.

Done.

gka (approved):
nice, thx!

gka:
actually, a tiny request: we should change mentions of "Chart" in the translations to "visualiztion", since this is also being used for maps and tables. can you do that @sjockers ?

sjockers:
@gka actually, I couldn't find any mentions of 'chart' in the new UI strings. Do you have anything specific in mind? 

For the German translation I'll replace "Diagramm" with "Visualisierung".

gka:
I've probably just seen it pop up on neurath which was perhaps running an older commit. good to go

gka:
or: I misplaced my comment, which was meant for https://github.com/datawrapper/controls/pull/190 :joy:

gka (approved):
looks good to go! thanks for the PR

PR #490, opened 2021-09-01
ui: mock navbar from new frontend

gka:
just some css and minor PHP changes to make old navbar look a bit like the new one


davidw-dw:
@gka we probably can delete this pr?

gka:
you can ignore the entire repo

gka:
the only reason it's not archived yet is that we intend to use it for hosting our "open" component code in the future

